PR #3, opened 2022-05-31
Adds a case for error "Type 'key' cannot be used as an index type." in Typing wiki

mridul-netizen:
Adds a case for error "Type 'key' cannot be used as an index type." in Typing wiki


mridul-netizen:
@seanlip PTAL

seanlip (approved):
LGTM. Thanks!

PR #6, opened 2022-06-12
Updated Guide on defining types (markdown)

JeeveshGarg:
Add one more case for delete operator.

### DIFF
--- a/Guide-on-defining-types.md
+++ b/Guide-on-defining-types.md
@@ -754,7 +754,7 @@ fingerprintHashesIndex.forEach((hashIndex: number) => {
  The operand of a 'delete' operator must be optional.
 #### [TS-9-1]:
 ##### Violation:
-This error occurs when using the delete operator in strictNullChecks. Because the property is no longer implemented after deletion, set its type to undefined. As a result, the property must be undefined; otherwise, the delete operator leads to an error.
+This error occurs when using the `delete` operator with strictNullChecks. Since a property cannot be implemented after deletion, its type is set to undefined. So, the property must contain undefined as a type, otherwise the delete operator leads to an error.
 
 ```typescript
 export class OutcomeDestinationEditorComponent implements OnInit {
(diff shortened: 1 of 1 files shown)


JeeveshGarg:
@mridul-netizen @vojtechjelinek PTAL.
Thanks.

vojtechjelinek (approved):
LGTM!

Guide-on-defining-types.md
```diff
@@ -749,3 +749,39 @@ fingerprintHashesIndex.forEach((hashIndex: number) => {
       fingerprint.push([kGramHashes[hashIndex], hashIndex]);
 });
 ```
+
+### [TS-9]:
+ The operand of a 'delete' operator must be optional.
+#### [TS-9-1]:
+##### Violation:
+This error occurs when using the delete operator in strictNullChecks. Because the property is no longer implemented after deletion, set its type to undefined. As a result, the property must be undefined; otherwise, the delete operator leads to an error.
```
mridul-netizen:
```suggestion
This error occurs when using the `delete` operator with strictNullChecks. Since a property cannot be implemented after deletion, its type is set to undefined. So, the property must contain undefined as a type, otherwise the delete operator leads to an error.
```

Guide-on-defining-types.md
```diff
@@ -749,3 +749,39 @@ fingerprintHashesIndex.forEach((hashIndex: number) => {
...
+  delete this.outcomeNewStateName; // error
+  ...
+}
+```
+
+
+##### Solution:
+One of the potential fix to this error is to assign `undefined` but generally we try to avoid unnecessary use of `undefined`. Because usage of `undefined` does not represent or give proper information about the actual situation.
+
+As a result, in this case, we choose to use `null` rather than `undefined`. Because the delete operator makes a property unusable, `null` is a better representation of emptiness, unusable or property is now no longer implemented.
```
mridul-netizen:
```suggestion
One of the potential fixes to this error is to assign `undefined` as a type but we generally try to avoid it's usage because it does not represent or give proper information about the actual situation.

So, in this case, we choose to use `null` rather than `undefined`. Since the delete operator makes a property unusable, `null` is a better representation of emptiness or unusability and property is now no longer implemented.
```

mridul-netizen (changes requested):

Guide-on-defining-types.md
JeeveshGarg: Done

Guide-on-defining-types.md
JeeveshGarg: Done

mridul-netizen (approved):

JeeveshGarg:
@mridul-netizen please merge this PR, I'm not authorized to merge this pull request.

mridul-netizen:
@seanlip @gp201 can you guys merge this? Seems like we don't have the rights!

U8NWXD:
@mridul-netizen for future reference, wiki PRs can only be merged by the welfare team. See https://github.com/oppia/oppia/wiki/Wiki#opening-a-pull-request for a discussion of the PR workflow for this repo

PR #11, opened 2022-07-09
Added links of webdriverio files and renames class names.

ShivamJhaa:
This PR adds the link of files of webdriverio and replaced clasname 'protractor-test-*' to 'e2e-test-*'.


WebdriverIO.md
```diff
@@ -186,6 +189,8 @@ It might be tempting to use the `buttons[0]`, `buttons[last]`, and `buttons[n]`
 
 Except for cases where an element selector is crafted dynamically, all element selectors should be at the top of a utility file. There should be no element selectors in suite files.
 
+Note: We need to use `await` prefix while using `$$` selector in order to avoid flakiness of tests.
```
SAEb-ai: Also, mention the case where we don't need it.

WebdriverIO.md
```diff
@@ -154,17 +154,20 @@ If you use one of options 2-4, you should create a chain of element selectors wh
...
 
 Notice that the top of the chain, where we select Element A, uses method 1.
 
 Sometimes you need to distinguish between several different elements which all look the same to element selectors. You can iterate over all the elements to find the right one. For example, suppose we want to click on the button to open a topic, where the button text is the topic name. We could find the right button like this:
 
 ```js
-var buttons = $$('.webdriverio-test-button');
+var buttonsSelector = function() {
+  return $$('.e2e-test-button');
+}
```
SAEb-ai, line 167: Wouldn't it be good if you explain this too a bit in terms of flakiness like why we need a function for all selector when it's declared at the top?

SAEb-ai (changes requested):

U8NWXD (approved):
This LGTM aside from SAEb-ai's comments

U8NWXD:
@SAEb-ai remember to update the assignees. We don't have Oppiabot here, so the assignments don't get updated automatically

WebdriverIO.md
ShivamJhaa: Done.

WebdriverIO.md
ShivamJhaa: Done.

WebdriverIO.md
```diff
@@ -182,10 +185,27 @@ var openTopic = async function(topicName) {
 }
 ```
 
+As you may have noted above that `$$` selector is used by calling a function instead of getting directly declared as variable like `$`, the reason is we cannot use `$$` without `await` prefix as this leads to flakiness in test because it returns an unresolved promise, and we can use `await` only inside async function. So as we need to place the selectors at top we are creating a function instead and using `await` prefix wherever we are calling the function.
+
```
SAEb-ai: >and we can use `await` only inside async function

I am not able to understand this. I think we use await prefix inside async function only, right? For eg see openTopic is an async function only. Hence we are able to use await.

SAEb-ai (changes requested):

WebdriverIO.md
ShivamJhaa: Yes, but we need to place the selector at top, and there we cannot use await.

SAEb-ai: My comment was why we can't use `await $$selector` at the top instead of having a function for it and then calling it because theoritically both would mean the same. This is in reference to what you said above i.e we need an await for $$ selector.

SAEb-ai (changes requested):

WebdriverIO.md
ShivamJhaa: Is it fine now?

WebdriverIO.md
```diff
@@ -182,10 +185,27 @@ var openTopic = async function(topicName) {
 }
 ```
 
+As you may have noted above that `$$` selector is used by calling a function beacause we need to use `await` prefix with `$$` selector and we cannot place `await $$` directly at top as it will try to find the element as soon as the test will run, and this will cause the test to fail.
```
SAEb-ai: Can you please also add the consequences of this sentence? From this line, my understanding is that it's good only right if the all selector element is found first when the test will run. But my question is what's the difference between declaring it at the top using await and then accessing it using the defined variable in comparision to declaring a function at the top and then calling the function in the tests in order to access the element.

`it will try to find the element as soon as the test will run`

SAEb-ai (changes requested):

WebdriverIO.md
SAEb-ai: PTAL at https://github.com/oppia/oppia-web-developer-docs/pull/11#discussion_r922803180

WebdriverIO.md
ShivamJhaa: Updated the comment PTAL.

WebdriverIO.md
```diff
@@ -182,10 +185,27 @@ var openTopic = async function(topicName) {
 }
 ```
 
+As you may have noted above that `$$` selector is used by calling a function beacause we need to use `await` prefix with `$$` selector and we cannot place `await $$` directly at top as `await` can only be used inside `async` function and if we will convert the function to `async` we cannot use it as a constructor for defining page objects.
```
SAEb-ai:
```suggestion
As you may have noted above that `$$` selector is used by calling a function beacause we need to use `await` prefix with `$$` selector and we cannot place `await $$` directly at top as `await` can only be used inside `async` function and if we convert the function to `async` we cannot use it because of [this](https://stackoverflow.com/questions/51694221/not-a-constructor-with-async-function#:~:text=When%20you%20use%20the%20async,can't%20be%20a%20constructor) error.
```

SAEb-ai (changes requested):

WebdriverIO.md
ShivamJhaa: Done.

ShivamJhaa:
@SAEb-ai PTAL, thanks

SAEb-ai (approved):

SAEb-ai:
@U8NWXD Please merge this PR. The changes LGTM to me.

PR #16, opened 2022-08-05
Updated Backend type annotations Doc for the usage of assert keyword.

sahiljoster32:


Backend-Type-Annotations.md
```diff
@@ -109,5 +109,23 @@ from typing import (
 
 ```
 
+4. While narrowing down the `Optional[...]` types (eg. Optional[str]) always keep assert at your last option, because if assert is not used properly it can create abmbiguity among other developers. For proper usage of assert, you can refer following points:
```
U8NWXD, line 112: I'm seeing some typos and grammar issues, e.g.

```suggestion
4. While narrowing down `Optional[...]` types like `Optional[str]`, always keep assert as your last option because improper asserts can create ambiguity for other developers. To use asserts properly, please follow these points:
```

ditto elsewhere.

Backend-Type-Annotations.md
```diff
@@ -109,5 +109,23 @@ from typing import (
...
+                'value not found.'
+            )
+        y = x    # Here, x has Optional[str] type and we know that x cannot be None here, so to narrow down it's we used assertion here with comment.
+        # Ruling out the possibility of None for mypy type checking, because x is
+        # already validated above. 
+        assert x is not None
+        return y
+    ```
+    - Please do not use assert if you are not sure that value is going to be None or not, instead try to debug and find the answer for `why it is None here?` or `Is that value have already been validated in caller functions?`. If value have already been validated in caller functions then raise an proper exception in called function so that in future if exception is raised then we know that value cannot be None here.
+
```
U8NWXD, line 127: unnecessary extra blank lines

U8NWXD (changes requested):
I think the guidance here is a little confusing. First you say to only use asserts as a last resort, but then you give an example of where asserts are appropriate even when other approaches are possible (e.g. `cast`). It seems like your guidance really is "Use asserts only to tell mypy something we've already confirmed with earlier code"

Backend-Type-Annotations.md
```diff
@@ -109,5 +109,23 @@ from typing import (
...
+            raise Exception(
+                'value not found.'
+            )
+        y = x    # Here, x has Optional[str] type and we know that x cannot be None here, so to narrow down it's we used assertion here with comment.
+        # Ruling out the possibility of None for mypy type checking, because x is
+        # already validated above. 
+        assert x is not None
+        return y
+    ```
+    - Please do not use assert if you are not sure that value is going to be None or not, instead try to debug and find the answer for `why it is None here?` or `Is that value have already been validated in caller functions?`. If value have already been validated in caller functions then raise an proper exception in called function so that in future if exception is raised then we know that value cannot be None here.
```
U8NWXD, line 126: The code blocks here should be quotes instead since they aren't code or anything else where we're worried about a reader mixing up characters

U8NWXD (changes requested):

Backend-Type-Annotations.md
seanlip, line 126: "if you are not sure that value" --> "if you are not sure whether that value"

"If value have already been validated in caller functions then raise an proper exception in called function so that in future if exception is raised then we know that value cannot be None here." --> "If the value is supposed to have already been validated in the caller functions, then raise a proper exception in the called function, so that unsupported inputs  get handled correctly."

seanlip (changes requested):

seanlip:
@sahiljoster32 are you planning to submit this?

sahiljoster32:
As per the earlier offline discussion, this PR's content is going to be covered in the M2's Backend annotations doc PR. Thanks!

PR #17, opened 2022-08-10
Add a new common Beam error

vojtechjelinek:


Apache-Beam-Jobs.md
```diff
@@ -670,6 +671,28 @@ The code above throws this error `'_UnwindowedValues' object is not subscriptabl
     ))
 ```
 
+### `_namedptransform is not iterable` error
+
+This error sometimes happens when you forget to add a label for some operation
+(the strings of code before `>>`).  The solution is to add a label for all operations.
```
U8NWXD: In the wiki we keep whole paragraphs on one line so that diffs don't show a bunch of changes just becase the text was re-wrapped

Apache-Beam-Jobs.md
```diff
@@ -670,6 +671,28 @@ The code above throws this error `'_UnwindowedValues' object is not subscriptabl
...
+#### Example
+
+```python
+some_values = (
+    some_models
+    | beam.Values()
+)
+```
+The code above might return `'_namedptransform is not iterable` in the job output.
+We can fix this by adding an appropriate label.
```
U8NWXD: ditto. This will be made into a single paragraph

U8NWXD (changes requested):
just some comments about breaking lines

Apache-Beam-Jobs.md
vojtechjelinek: done

Apache-Beam-Jobs.md
vojtechjelinek: done

U8NWXD (approved):

PR #34, opened 2022-09-27
Instructions on how to create Rosetta terminal

MattiasErlingson:
Add instructions document on how to create a rosetta terminal on a Mac M1 computer and linked it to the installation instructions. From my own experience, much easier to have a dedicated terminal using rosetta along the entire installation process.


Creating-rosetta-terminal.md
```diff
@@ -0,0 +1,17 @@
+## How to create a Rosetta 2 terminal on Mac M1 computers
```
U8NWXD: Can we put these instructions in the mac installation instructions? Since they're so short, it seems unnecessary to have a whole separate page for them

U8NWXD (changes requested):

Creating-rosetta-terminal.md
```diff
@@ -0,0 +1,17 @@
...
+
+#### Install Rosetta 2
+Rosetta 2 is only available for Mac computers with the M1 chip. If Rosetta 2 is not installed, prompt a rosetta 2 installation using the command:
+
+`softwareupdate --install-rosetta`
+
+#### Create a Rosetta terminal
+
+1. Open Finder and navigate to Application/Utilities. 
+2. Right-click the "Terminal" application and select "Duplicate".
```
U8NWXD: This step seems very non-standard. I'm checking with some devs who use M1 macs to confirm that this is safe

U8NWXD: I haven't been able to find anyone else who has a good understanding of the potential side-effects of this, so let's add a disclaimer that this step might be dangerous and provide an alternative for users who don't want to take that risk

MattiasErlingson: Sure. Alternatively, we skip the "Duplicate" part and only instruct users to create a "session" terminal by clicking "open using Rosetta".

Creating-rosetta-terminal.md
MattiasErlingson: Yeah, guess the most reasonable place would be in step 1 in "Macs with M1 chips"?

MattiasErlingson:
Hi @U8NWXD, Found an alternative way of changing architecture for a session, that includes using aliases. This way, there's no need to duplicate the application. I put this in the Mac m1 installation guide and removed the separate document for it.

Installing-Oppia-(Mac-OS;-Python-3).md
```diff
@@ -19,11 +19,17 @@ To check whether your Mac has an M1 chip, navigate to the Apple menu and select
...
-1. [Install](https://stackoverflow.com/a/64883440) Rosetta 2
+1. If not yet installed, install Rosetta 2 with the following command:  `softwareupdate --install-rosetta`. Rosetta 2 translates Intel-based apps to run on Apple silicon Macs.
 
-2. Inside Rosetta perform the Downloading and prerequisites steps (**Note:** If `sudo easy_install pyyaml` does not work try using `pip3 install pyyaml`).
+2. Next, we'll create two aliases to make it more convenient to switch between Mac M1s' *ARM* architecture and the emulated *Intel* architecture. To do so, run the following in the terminal (change "zsh" to "bash" if you're using a bash terminal):
+    ```shell
+    $ alias arm="env /usr/bin/arch -arm64 /bin/zsh --login"
+    $ alias intel="env /usr/bin/arch -x86_64 /bin/zsh --login"
+    ``` 
+    Now, run `intel` in the terminal followed by `arch`. You should see `i386` being printed. This is the architecture we'll use for this installation process. (**Note:** To change back to ARM64, run "arm" or simply open a new terminal).
```
U8NWXD: Running `arm` here will open a new shell inside the existing shell, so it's not quite switching. This might be important because if users run arm/intel a lot, they'll make a lot of nested shells. You can see this by running `exit` after you call arm/intel a few times--you'll have to run `exit` multiple times to exit out of all the nested shells

U8NWXD (changes requested):
Sorry for the late review--I lost track of this PR. I like your solution! I just have one small comment

U8NWXD (approved):

PR #64, opened 2022-12-12
Update macOS Installation Instructions for direnv and Apple Silicon

### LINKED ISSUE #63: Update macOS installation instructions to use direnv
Requested by @vojtechjelinek in https://github.com/oppia/oppia-web-developer-docs/pull/5#issuecomment-1179334482. Direnv is helpful because it puts the virtual environment inside the `opensource/` directory where it's easier to find than in `~/.pyenv`.

U8NWXD:
Add instructions on using `direnv` and fixes for Macs that use Apple Silicon. Fixes #63.

### DIFF
--- a/Installing-Oppia-(Mac-OS;-Python-3).md
+++ b/Installing-Oppia-(Mac-OS;-Python-3).md
@@ -9,7 +9,7 @@
 
 **Note:** If you just want to create and share explorations, you may be able to use the hosted server at https://www.oppia.org (in which case you don't need to install anything).
 
-*These installation instructions were last tested on 24 July 2021. For more information on issues that may occasionally arise with the installation process, please see the [Troubleshooting](https://github.com/oppia/oppia/wiki/Troubleshooting) page.*
+*These installation instructions were last tested on 11 Dec 2022. For more information on issues that may occasionally arise with the installation process, please see the [Troubleshooting](https://github.com/oppia/oppia/wiki/Troubleshooting) page.*
 
 **Note:** Be careful about trying to install Oppia if you have the Python [Anaconda platform](https://www.anaconda.com/) installed. We've received a bunch of reports that installation is tricky in that environment (there are lots of small things that get in the way), and that the solution is to use the standard python installation (via e.g. homebrew) instead.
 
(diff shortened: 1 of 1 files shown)


Installing-Oppia-(Mac-OS;-Python-3).md
```diff
@@ -9,62 +9,81 @@
...
 
-2. Ensure that you have [Python 3.8](https://www.python.org/downloads/release/python-3812/) installed (Note: you can check this by running `python3 --version`). If Python 3.8 is not installed, download and run the latest Python 3.8 installer from https://www.python.org/downloads/mac-osx/.
+## Install prerequisites
+
+Oppia relies on a number of programs and third-party libraries. Many of these libraries are downloaded automatically for you when you first run the `start.py` script provided with Oppia. However, there are some things that you will need to do beforehand:
+
+1. Install [version 8+ of the Java Runtime Environment (JRE)](https://www.java.com/en/download/).
 
-3. Download [git](http://git-scm.com/download/mac), then run the package and follow instructions. This allows you to store the source in version control.
+2. Install [direnv](https://direnv.net/).
```
gp201, line 69: I believe the cmd is `brew install direnv`. Can we put that instead of saying install direnv?

Installing-Oppia-(Mac-OS;-Python-3).md
```diff
@@ -9,62 +9,81 @@
...
+
+   After finishing with the official installation instructions, you should also create a `~/.direnvrc` file with the following content:
```
gp201, line 71: Make this a new step

Installing-Oppia-(Mac-OS;-Python-3).md
```diff
@@ -9,62 +9,81 @@
...
+       layout python "$python_root/bin/python"
+     else
+       echo "Error: $python_root/bin/python can't be executed."
+       exit
+     fi
+   }
+   ```
 
-**Note:** While #1 above is only for those having MacOS >= 12.3, #2 and #3 should be done by everyone using a Mac for setting up Oppia. 
+3. Download [git](http://git-scm.com/download/mac), then run the package and follow instructions. This allows you to store the source in version control. Note that you can also install Apple's command line tools to get git instead.
```
gp201, line 86: Can we put `brew install git` to ensure everyone is installing from the same source? It will be easier to debug if devs face an issue with git.

Installing-Oppia-(Mac-OS;-Python-3).md
```diff
@@ -150,69 +169,56 @@ For your vitual environment, we recommend you use [pyenv](https://github.com/pye
...
    ```console
-   pyenv local oppia
+   pyenv global 3.8.12 2.7.18
    ```
 
-   Now whenever you are within the `oppia` folder, the virtual environment will be active.
+   Note that you can use a different version of Python 3 if you prefer--we'll specify Python 3.8.12 as our version for Oppia development later.
 
-5. Install the Python dependencies:
+5. Create a virtual environment for oppia by making a file `opensource/.envrc` with the following contents:
```
gp201: Users may assume that `opensource/.envrc` is the file name. Can you specify the folder and file name separately?

Installing-Oppia-(Mac-OS;-Python-3).md
```diff
@@ -150,69 +169,56 @@ For your vitual environment, we recommend you use [pyenv](https://github.com/pye
...
-     Downloading PyYAML-5.4.1-cp27-cp27mu-manylinux1_x86_64.whl (574 kB)
-        |████████████████████████████████| 574 kB 2.3 MB/s
-   Installing collected packages: pyyaml
-   Successfully installed pyyaml-5.4.1
+   ```text
+   use python 3.8.12
    ```
 
-6. If you want to run backend tests and check coverage, please install these 2 pip libraries globally (or in your venv).
+   Then run `direnv allow` in `opensource/` to allow `direnv` to run there. Whenever you are within the `opensource/` directory (or any of its subdirectories, or any of their subdirectories, etc.) the virtual environment will be active. You can confirm this by running:
```
gp201:
```suggestion
   Then run `direnv allow` in the `opensource/` directory to allow `direnv` to run there. Whenever you are within the `opensource/` directory (or any of its subdirectories, or any of their subdirectories, etc.) the virtual environment will be active. You can confirm this by running:
```

gp201 (changes requested):

Installing-Oppia-(Mac-OS;-Python-3).md
U8NWXD: There are more steps than just `brew install direnv`, for example changes to *rc files to hook direnv into the user's shell. I link to the direnv instructions because they do a good job of explaining what is required, will be kept up to date if newer versions of direnv need different *rc file changes, and cover other shells besides bash (e.g. zsh, fish)

Also, there's currently no requirement that devs use Homebrew to install their dependencies, so I don't want to require that they install direnv from Homebrew

Installing-Oppia-(Mac-OS;-Python-3).md
U8NWXD: done

We assume throughout our installation instructions that users understand path notation. Do you think we need to re-evaluate that assumption?

Installing-Oppia-(Mac-OS;-Python-3).md
U8NWXD: I don't think I understand. Why does `brew install git` result in everyone installing from the same source, while `install from git-scm.com` doesn't? If your point is just that we should give one source instead of the note about Apple's command line tools, then why install from Homebrew over the two sources currently listed?

I actually think installing from Apple's command-line tools makes the most sense. Users will probably have to install those anyway, and there's no reason to have multiple versions of git on your system at once. wdyt?

Also, there's currently no requirement that devs use Homebrew to install their dependencies

Installing-Oppia-(Mac-OS;-Python-3).md
U8NWXD: I disagree. Right now, we have one bullet point for each dependency, which conceptually makes sense. If we made this a new bullet point, then direnv would get 2 steps unlike the other dependencies. I'll update the text above to make clear that each point is a dependency, not a step

Installing-Oppia-(Mac-OS;-Python-3).md
U8NWXD: done

Installing-Oppia-(Mac-OS;-Python-3).md
gp201: I believe it would be more intuitive to list each step as an action a user has to take rather than a dependency (Something like what we have for the Linux installation). Users tend to glance over the steps instead of reading each step completely; this may cause issues with their setup.

Installing-Oppia-(Mac-OS;-Python-3).md
gp201: My bad, you're right. However, since this is for Mac OS only, I think it would be easier for the devs to simply copy the command directly from our wiki rather than going to a new website and figuring out how they want to install it.

Installing-Oppia-(Mac-OS;-Python-3).md
gp201: Yes, I have observed a few devs assuming that to be the filename itself and not the path/filename.

Installing-Oppia-(Mac-OS;-Python-3).md
gp201: Sounds good!

probably add a comment (if you can) to the wiki.

gp201 (changes requested):
Thanks @U8NWXD I replied to your comments

Installing-Oppia-(Mac-OS;-Python-3).md
U8NWXD: I'm still unclear on what approach you think is best. Here are the options:

1. Install from git-scm.com. This can't be done just by running a terminal command because the installer has a GUI. Also, we want people to get the latest version, but we would have to pick a version to give a `curl` or `wget` download command.
2. Install from Homebrew. This requires that they install Homebrew, so it's also not as simple as copying a command from the wiki. They would need to (a) run the Homebrew installation command, (b) modify the *rc file for their shell, and (c) install git.
3. Install Apple's command-line tools. This can be started from the command-line, but the installer is again graphical, so the command-line is insufficient.

Installing-Oppia-(Mac-OS;-Python-3).md
U8NWXD: We also group related steps in the Linux instructions. For example, here's step 8 of the [virtual environment setup](https://github.com/oppia/oppia/wiki/Installing-Oppia-%28Linux%3B-Python-3%29#setup-a-virtual-environment):

> Create a virtual environment for oppia by adding file named .envrc into the parent folder of the oppia repository with this content:
> 
> ```console
> use python 3.8.12
> ```
> 
> Then run this command in the same folder:
> 
> ```console
> $ direnv allow
> ```
> 
> Now whenever you are within the oppia folder, the virtual environment will be active.

I think it's the contributor's responsibility to read the instructions thoroughly. If they can't do that, they're going to have a very hard time with the rest of the development cycle, e.g. reading our more complex wiki pages like on debugging. This is especially relevant since we're in the process of raising the bar to contributing.

The benefit of breaking the installation process into steps is that it creates conceptual chunks to organize the information, which is best served by keeping conceptually-related information together. Splitting the process into too many steps is just as confusing as having too few. In the direnv case, if we split this into 2 steps, then we would have 4 steps: 2 of them would tell the user "here's how to install x" while the other two would describe "one of the two steps required to install x". I think that inconsistency is confusing. Instead, I'd rather have 3 steps, each of which say "here's how to install x".

gp201: I agree with you that its the contributor's responsibility to read the instructions thoroughly. However, this may backfire on us if multiple contributors start asking the same question. For now, let's proceed with the existing format it's written in and change it if we start seeing multiple contributors getting suck in that step. WDYT?

Installing-Oppia-(Mac-OS;-Python-3).md
gp201: Hmmm, installing from the cmd is more tedious than I had anticipated. But thinking again, contributors should be able to install git on their own. Hence let's go with the original text you had written.

gp201 (approved):
LGTM. replied to your comments.

Installing-Oppia-(Mac-OS;-Python-3).md
U8NWXD: sg

PR #76, opened 2023-01-02
Point new contributors to the actual good first issues rather than the list on the page which is getting stale.

seanlip:
I think new contributors might be having trouble finding good first issues to tackle, and this might be because they are slightly hidden in the new project boards. This change makes the list of good first issues more prominent and removes the hardcoded list on the wiki page that is getting stale.

### DIFF
--- a/Contributing-code-to-Oppia.md
+++ b/Contributing-code-to-Oppia.md
@@ -25,9 +25,11 @@ If you run into any problems along the way, we're here to help! Check out our [[
 
 2. Fill in the [Oppia contributor survey](https://goo.gl/forms/otv30JV3Ihv0dT3C3) to let us know what your interests are. (You can always change your responses later.)
 
-3. Install Oppia following the [[installation instructions|Installing-Oppia]]. If you run into any issues, please check out the [[troubleshooting instructions|Troubleshooting]]. If you want help setting up a code editor, also check out our [[guide to common IDEs|Tips-for-common-IDEs]]. If the above resources don't help and you're still stuck, please check [GitHub Discussions](https://github.com/oppia/oppia/discussions/categories/q-a-installation) to see if any existing threads address the issue. If not, feel free to start a new thread explaining what you've tried and what you're seeing, so that we can try and help you!
+3. Say hi and introduce yourself on [GitHub Discussions](https://github.com/oppia/oppia/discussions/16715)!
 
-4. Update your GitHub settings:
+4. Install Oppia following the [[installation instructions|Installing-Oppia]]. If you run into any issues, please check out the [[troubleshooting instructions|Troubleshooting]]. If you want help setting up a code editor, also check out our [[guide to common IDEs|Tips-for-common-IDEs]]. If the above resources don't help and you're still stuck, please check [GitHub Discussions](https://github.com/oppia/oppia/discussions/categories/q-a-installation) to see if any existing threads address the issue. If not, feel free to start a new thread explaining what you've tried and what you're seeing, so that we can try and help you!
+
+5. Update your GitHub settings:
 
    * [Set up 2FA]( https://help.github.com/articles/securing-your-account-with-two-factor-authentication-2fa/) on your GitHub account. **This is important to prevent people from impersonating you.**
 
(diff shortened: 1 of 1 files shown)


seanlip:
@gp201 @JeeveshGarg PTAL at these updates to the onboarding wiki. Thanks!

Contributing-code-to-Oppia.md
```diff
@@ -61,37 +61,27 @@ We also **strongly recommend** looking through the resources under "Developing O
...
+### Choosing a good first issue
 
-* [#10616](https://github.com/oppia/oppia/issues/10616) (Introduce new @typescript-eslint rules in the codebase)
+When you've found a good first issue you'd like to tackle, please leave a comment on it that:
 
-These issues are hand-picked to ensure that you don't run into unexpected roadblocks while working on them. For other issues, you may need to be more independent because often times, we don’t know how to solve them either.
+- @-mentions the lead(s) of the corresponding project that the issue falls under (see details for [LaCE](https://github.com/orgs/oppia/projects/3/views/8?pane=info), [Contributor Dashboard](https://github.com/orgs/oppia/projects/18/views/4?pane=info) or [Developer Workflow](https://github.com/orgs/oppia/projects/8?pane=info)), letting them know you'd like to work on it.
+- says which part of the issue you're taking, if the issue is divided into subparts, and
+- describes in more detail how you'd tackle the issue (e.g. explain which file(s) you would modify and what changes you would make). If your explanation makes sense, we'll assign the issue to you.
```
JeeveshGarg, line 80: can we add one more point. Example see previous merged PRs, because they are great source of learning.  As i also learn too much from that. Your thoughts?

JeeveshGarg (commented):
@seanlip PTAL!

Contributing-code-to-Oppia.md
seanlip: Thanks for taking a look!

Hm, in this case I lean towards not adding it. here This is because that point is mainly relevant for PRs that are "checkbox-like", and in such cases we should just include that info in the issue itself.

For other types of issues (which are going to be the majority going forward), I think it would not be straightforward for folks to figure out which PRs are a good source of learning, so this will just raise more questions.

For now, I think that we can monitor the explanations folks give when they want to tackle an issue. If we see gaps there that would be solved by looking at previous PRs, we could update this to include more specific instructions for how to do that.

How does this sound to you?

JeeveshGarg:
@seanlip left one comment.

seanlip:
@JeeveshGarg Thanks, already replied. PTAL.

Contributing-code-to-Oppia.md
JeeveshGarg: Yup sound good to me. It make sense. Will add advice to issues description, in which checkbox is there.

JeeveshGarg:
@seanlip LGTM! Thanks.

seanlip:
@gp201 doesn't seem to be around and I still see requests coming in from contributors, so I am going to go ahead and merge this since Jeevesh has approved it.

Contributing-code-to-Oppia.md
```diff
@@ -61,37 +61,27 @@ We also **strongly recommend** looking through the resources under "Developing O
...
 
-* [#8038](https://github.com/oppia/oppia/issues/8038) (Refactoring frontend services - The backend api calls in the frontend should only happen through services)
+After that, you can choose a good first issue from the [list of good first issues](https://github.com/oppia/oppia/labels/good%20first%20issue). These issues are hand-picked to ensure that you don't run into unexpected roadblocks while working on them, and each of them should have clear instructions for new contributors. If you see one that doesn't, please let us know via [GitHub Discussions](https://github.com/oppia/oppia/discussions) and we'll fix it. For other issues, you might need to be more independent because we might not know how to solve them either.
 
-* [#10700](https://github.com/oppia/oppia/issues/10700) (Refactor all migrated object factories to model.ts files in angular)
+You can also browse good first issues for each of the core Oppia Web teams to find something you'd enjoy working on:
 
-* [#8668](https://github.com/oppia/oppia/issues/8668) (Documenting the services)
+- Learner and Creator Experience (LaCE): https://github.com/orgs/oppia/projects/3/views/10, typically frontend or full-stack
```
gp201, line 70: Can you convert `Learner and Creator Experience (LaCE` into a hyperlink instead of showing the URL?

Contributing-code-to-Oppia.md
```diff
@@ -61,37 +61,27 @@ We also **strongly recommend** looking through the resources under "Developing O
...
+- Contributor Dashboard: https://github.com/orgs/oppia/projects/18/views/5, typically full-stack
+- Developer Workflow: https://github.com/orgs/oppia/projects/8/views/10, typically backend or frontend
 
-* [#10474](https://github.com/oppia/oppia/issues/10474) (Make typescript checks strict)
+### Choosing a good first issue
```
gp201, line 74: Please update the ToC

gp201 (commented):
@seanlip I left a few comments. Can you please address these in a new PR?

PR #84, opened 2023-01-14
Check Note and warning notations

### LINKED ISSUE #29: Change "Note" and "Warning" Notations to be more noticeable.
Move all instances of:

**Note:** While #1 above is only for those having MacOS >= 12.3, #2 and #3 should be done by everyone using a Mac for setting up Oppia.
to 
> **Note**
> While #1 above is only for those having MacOS >= 12.3, #2 and #3 should be done by everyone using a Mac for setting up Oppia.

```
> **Note**
> While #1 above is only for those having MacOS >= 12.3, #2 and #3 should be done by everyone using a Mac for setting up Oppia.
```

and

Warning: {text}
to 
> **Warning**  
> {text}

```
> **Warning**  
> {text}
```

SD-13:
Fixes #29 
Add lint check for **Note** and **Warning** notations.


SD-13:
Hi @gp201 , Please take a pass when you are free. Thanks,

U8NWXD:
@SD-13 is it possible to avoid having a third-party dependency for this? If not, we need to thoroughly vet the dependency to make sure that it's developer is trustworthy, that it's well-maintained, and that it introduces no security concerns.

I took a brief look, and I'm a little worried because avtodev/markdown-lint is just the start in a chain of dependencies. [avto-dev/markdown-lint](https://github.com/avto-dev/markdown-lint) appears to mostly be just a Dockerized version of [igorshubovych/markdownlint-cli](https://github.com/igorshubovych/markdownlint-cli), which in turn is just a wrapper for [DavidAnson/markdownlint](https://github.com/DavidAnson/markdownlint). This PR requires that we trust all 3 projects.

Is it possible to use markdownlint directly? That would simplify the vetting process and reduce our attack surface

SD-13:
> @SD-13 is it possible to avoid having a third-party dependency for this? If not, we need to thoroughly vet the dependency to make sure that it's developer is trustworthy, that it's well-maintained, and that it introduces no security concerns.
> 
> I took a brief look, and I'm a little worried because avtodev/markdown-lint is just the start in a chain of dependencies. [avto-dev/markdown-lint](https://github.com/avto-dev/markdown-lint) appears to mostly be just a Dockerized version of [igorshubovych/markdownlint-cli](https://github.com/igorshubovych/markdownlint-cli), which in turn is just a wrapper for [DavidAnson/markdownlint](https://github.com/DavidAnson/markdownlint). This PR requires that we trust all 3 projects.
> 
> Is it possible to use markdownlint directly? That would simplify the vetting process and reduce our attack surface

Okay got you. I am trying to use a bash script for this. We can use that right?

U8NWXD:
Writing our own scripts is fine as far as the security implications of third-party dependencies go. If the script is complicated, I might prefer Python since bash logic can be hard for devs used to Python/JS to read. For simple scripts though, bash is fine

gp201:
@SD-13 I have converted this PR to a draft. Once you have addressed https://github.com/oppia/oppia-web-developer-docs/pull/84#issuecomment-1383051132 please reopen the PR for review. Thanks!

SD-13:
Hey @U8NWXD, I think in this case(for checking notations) it is possible to do this without using a third-party dependency. Having said that, I tried to add a lint check in python (for the reason you have already mentioned) instead of using markdownlint. For more context about how it will look like, you can see [this](https://github.com/SD-13/oppia-web-developer-docs/actions/runs/3976306554/jobs/6816703334) workflow. Please take a look when you are free and let me know if you have any suggestions. Thanks!

linter.py
```diff
@@ -0,0 +1,39 @@
+import os
```
U8NWXD: Can you create a `scripts` folder to put this linter in? That will help keep the code files organized

linter.py
```diff
@@ -0,0 +1,39 @@
...
+import re
+from typing import Pattern, List
```
U8NWXD: Thanks for including type hints! We should probably include mypy in our CI checks. Can you open an issue for that so we don't forget?

linter.py
```diff
@@ -0,0 +1,39 @@
...
+def find_regex_in_file_content(
+    search_pattern: Pattern,
+    correct_pattern: Pattern,
+    filenames: List[str]
+) -> bool:
+    """Function that looks for a regex pattern in the given file contents."""
+
+    exit_code = 0
+    for filename in filenames:
+        with open(filename) as file:
```
U8NWXD: Please add the reading mode and encoding, for example:

```suggestion
        with open(filename, 'r', encoding='utf-8') as file:
```

linter.py
```diff
@@ -0,0 +1,39 @@
...
+            lines = file.readlines()
+        for i, line in enumerate(lines):
```
U8NWXD: You can also just do `for i, line in enumerate(file)`, which is slightly more memory-efficient because you only load one line at a time in to memory instead of loading the whole file.

linter.py
```diff
@@ -0,0 +1,39 @@
...
+            if search_pattern.search(line):
+                print(f'{filename}:{i+1}: error: wrong notation "{search_pattern.pattern}"; expected "{correct_pattern.pattern}"')
```
U8NWXD: I think it's confusing to show the user a regex instead of the string that matches the regex. Also, I think instead of `correct_pattern` you can just have an example of the correct string (since you never actually use it as a regex)

linter.py
```diff
@@ -0,0 +1,39 @@
...
+                exit_code = 1
```
U8NWXD: Could this be `match_found` instead since we don't actually use it as an exit code?

linter.py
```diff
@@ -0,0 +1,39 @@
...
+found_note_pattern: bool = find_regex_in_file_content(
+    search_note_pattern, correct_note_pattern, all_markdown_filenames
+)
+found_warning_pattern: bool = find_regex_in_file_content(
+    search_warning_pattern, correct_warning_pattern, all_markdown_filenames
+)
+
+if found_note_pattern or found_warning_pattern:
+    print("Lint Check Failed!!")
+    exit(1)
```
U8NWXD: Please put these into a `main()` function that gets called under `if __name__ == '__main__'` so that this code doesn't run when the file is only imported

Instructions-for-Reviewers.md
```diff
@@ -35,7 +35,8 @@ Simply visit this URL: https://github.com/pulls/assigned. It takes you to a list
      * **Important:** Please also ensure that, if a PR author files a TODO for future work, they assign themselves to it (or arrange for it to be assigned responsibly to someone else) and don't just abandon it. It should typically be the next thing they work on, unless there is a principled reason to agree otherwise. (In other words, PR authors shouldn't just file TODOs for future work and then abandon them.)
 
 
-      **Note**: To checkout branch BRANCH_NAME from committer COMMITTER_USERNAME to your local machine, run:
+      > **Note**
+      > To checkout branch BRANCH_NAME from committer COMMITTER_USERNAME to your local machine, run:
       ```
```
U8NWXD: I think the lines below also go inside the note, right?

.github/workflows/linter.yml
```diff
@@ -0,0 +1,14 @@
+name: Lint Check
+
+on: push
```
U8NWXD: Can we make this run on PRs to `develop` too?

U8NWXD (changes requested):

U8NWXD:
Thanks @SD-13 this looks great! I left a few notes

linter.py
SD-13: Yep, I created https://github.com/oppia/oppia-web-developer-docs/issues/97 for this. PTAL

.github/workflows/linter.yml
SD-13: sure. and should we restrict the push only to develop?

linter.py
SD-13: done

linter.py
SD-13: added

linter.py
SD-13: updated

linter.py
SD-13: Yep, that makes sense. I have updated the output message regex with strings.

linter.py
SD-13: Yeah, `match_found` sounds self-explanatory. I have updated the code accordingly

linter.py
SD-13: Thanks for the explanation. Updated,

Instructions-for-Reviewers.md
SD-13: yes, you are right. Updated,

SD-13:
Thanks @U8NWXD I have addressed all your comments and added your suggested changes. PTAL,

scripts/linter.py
```diff
@@ -0,0 +1,44 @@
+import os
+import re
+from typing import Pattern, List
+
+search_note_pattern: Pattern = re.compile(r'^\*\*Note\*\*')
+search_warning_pattern: Pattern = re.compile(r'^\*\*Warning\*\*')
+correct_note_string: str = '> **Note**'
+correct_warning_string: str = '> **Warning**'
```
U8NWXD: Since these are constants, can you make them all-caps?

scripts/linter.py
```diff
@@ -0,0 +1,44 @@
...
+        search_warning_pattern, correct_warning_string, all_markdown_filenames
+    )
+
+    if found_note_pattern or found_warning_pattern:
+        print("Lint Check Failed!!")
+        exit(1)
+
+
+if __name__ == '__main__':
+    main()
```
U8NWXD: Please end the file with a newline

scripts/linter.py
```diff
@@ -0,0 +1,44 @@
...
+    ]
+    found_note_pattern: bool = find_regex_in_file_content(
+        search_note_pattern, correct_note_string, all_markdown_filenames
+    )
+    found_warning_pattern: bool = find_regex_in_file_content(
+        search_warning_pattern, correct_warning_string, all_markdown_filenames
+    )
+
+    if found_note_pattern or found_warning_pattern:
+        print("Lint Check Failed!!")
```
U8NWXD:
```suggestion
        print("Lint Check Failed!")
```

Instructions-for-Reviewers.md
```diff
@@ -35,12 +35,13 @@ Simply visit this URL: https://github.com/pulls/assigned. It takes you to a list
...
         git checkout -b {{COMMITTER_NAME}}-{{BRANCH_NAME}} develop
         git pull https://github.com/{{COMMITTER_NAME}}/oppia.git {{BRANCH_NAME}}
       ```
```
U8NWXD: This code block goes inside the note too. Right now it's not:
Screenshot 2023-01-24 at 20 58 34

U8NWXD (changes requested):

scripts/linter.py
SD-13: Done.

scripts/linter.py
SD-13: added new line

scripts/linter.py
SD-13: done

Instructions-for-Reviewers.md
SD-13: My bad. Last time I didn't get this but fixed this time.

SD-13:
Hey @U8NWXD, I have pushed all your suggested changes. Please let me know if you have anything else for me here. Thanks,

U8NWXD (approved):

PR #86, opened 2023-01-16
Improvements to the wiki article on 'Feature Gating'

nithinrdy:
1. Fixes example showing how to add a new feature flag to the `PARAM_NAMES` enum.
2. Fixes example showing how to register a feature flag.
3. Adds a section explaining how to modify unit tests to account for the newly created flag and prevent them from failing.
4. Adds instructions on writing front-end tests for gated features.
5. Adds instructions on writing back-end tests for gated features.
6. Elaborates on what a 'Feature Stage' means and how enabling/disabling a flag works.
7. Fixes some of the links pointing to headers.
8. Adds a section on why we need feature flags.
9. Adds a section on how to use them (sub-sections include when to use them, and how to use them).

Other minor rewording.


nithinrdy:
@seanlip PTAL.

Developing-new-features-with-feature-gating.md
```diff
@@ -60,10 +99,10 @@ from core.domain import platform_feature_services
 # ...
 
 if platform_feature_services.is_feature_enabled(
-    	platform_feature_list.PARAM_NAMES.dummy_feature):
+      platform_feature_list.ParamNames.DUMMY_FEATURE.value):
```
seanlip: Should this be deindented by 2?

seanlip (changes requested):
Thanks @nithinrdy! Looks great, just one minor inline comment. 

One more general comment though. I wonder whether we can make this page more explicit on "how we should use feature flags", not just "how feature flags work". I.e. can we have instructions that are like, "when developing a new feature, do XYZ, and then when it passes product review, do ABC, and then when ..."? That way, developers know exactly what to do and who to talk to at each stage of the process, and then they can own the launch of their own feature with confidence.

Wdyt?

nithinrdy:
> I wonder whether we can make this page more explicit on "how we should use feature flags", not just "how feature flags work". I.e. can we have instructions that are like, "when developing a new feature, do XYZ, and then when it passes product review, do ABC, and then when ..."? That way, developers know exactly what to do and who to talk to at each stage of the process, and then they can own the launch of their own feature with confidence.
> 
> Wdyt?


Sorry for the delay; that sounds good, I'll add a "Why do we need feature flags?" section explaining how they're used during a feature's development cycle.

seanlip:
Thanks @nithinrdy! Just a note though -- the intent is not just to have an info section about "why do we need feature flags", it's meant to be more of a recipe in terms of "please do X Y Z in order at each of these stages when you want to launch a new feature". I.e. it should be fairly directive, so that folks know that they should follow it and it's not just a suggestion. Think that would be fine?

Developing-new-features-with-feature-gating.md
```diff
@@ -93,9 +132,37 @@ We use the following principles to determine the value of a feature flags:
...
+
+1. The very first PR you make must include changes that add a new feature flag to the codebase, meant to be used with this new feature. The feature flag must be placed in the DEV stage. Every single user-facing aspect of the feature -- whether frontend or backend -- must be gated behind the feature flag (both in the first PR, and all the following ones as well). This is to ensure that the feature is not visible to the user until it is fully implemented and ready for production.
+
+2. The first PR must be merged before any following PRs are merged. This is to ensure that the feature flag is available in the codebase for it to be used in the following PRs.
+
+3. The very last PR you make (to finish up the feature you are working on) must include changes that move the feature flag to the TEST stage. This is to ensure that the feature is available in the test environment, and we can feature-test it before it is made available to the users in the production environment.
+
+4. If the feature testing reveals that the feature is not ready for production, you must work on fixing the highlighted issues before proceeding further.
```
nithinrdy: In this step, we could instruct developers to move the flag back to the DEV stage while they work on fixing the bugs, but that isn't _really_ necessary as the feature still cannot be enabled in prod anyway, and having to move the flag back (and then to TEST again for testing) will only add to the complexity of the process. It's a trade-off. Thoughts @seanlip?

nithinrdy:
@seanlip PTAL

Developing-new-features-with-feature-gating.md
```diff
@@ -41,6 +44,40 @@ export enum FeatureNames {
 }
 ```
 
+### Modifying unit tests to account for the creation of the new flag
```
seanlip, line 47: I'm wondering a couple of things:

- What is the desired practice for unit tests -- should they test both the case when the flag is off and when the flag is on? If so, maybe add instructions on how to turn a flag on in the tests.
- If this section isn't for the previous point, then what is it for exactly? I'm not quite sure; more explanation might be helpful.

Developing-new-features-with-feature-gating.md
```diff
@@ -49,6 +86,8 @@ Features fall in the three types of stages: dev, test and prod. In short, the st
 - test feature can be enabled in dev or test environments, but it can never be enabled in production environments.
 - prod feature can be enabled in any of the dev, test, production environments.
 
+Note: The environment the flag is placed in determines the enviroment(s) where it CAN BE enabled, not where it IS enabled. A feature in `test` stage, as mentioned above, can only be enabled in the `dev` and `test` environments, and not in the `prod` environment. By default, however, it is disabled in all environments. See the [Changing Value of Feature Flags](#changing-value-of-feature-flags) section for details on how to enable/disable feature flags.
```
seanlip: When you say "it is disabled in all environments", what is "it" and what does the "default" state refer to exactly?

Developing-new-features-with-feature-gating.md
```diff
@@ -93,9 +132,37 @@ We use the following principles to determine the value of a feature flags:
...
 - Each feature flag has multiple *rules*, its value changes to the value specified in the first rule that is matched. If no rule is matched, its value remains default.
 - Each rule has multiple *filters*, and a rule is matched only when all of its filters are matched.
-- Each filter describes a scenario, there are multiple types of filters like `server mode`, `client type`, so that you can you can describe scenarios like `server mode = dev OR server mode = test`.
-- Each filter can have multiple conditions, and a filter is matched if any of the conditions are matched.
+- Each filter describes a scenario. There are multiple types of filters (like `server mode` and `client type`) which can be used in combination with each other to describe various scenarios (like `server mode = dev OR server mode = test`).
+- Each filter can have multiple conditions, and a filter is matched if any of the conditions specified in the rule are matched.
+
+For example, if we want a feature flag to be enabled only in dev environments, we can configure the flag's rules/filters as shown below.
+
+![Example feature setting](https://i.imgur.com/GnFQ7El.jpg)
```
seanlip: Can you host this and any other images locally on the wiki (see other pages)?

Developing-new-features-with-feature-gating.md
```diff
@@ -93,9 +132,37 @@ We use the following principles to determine the value of a feature flags:
...
+
+Feature flags vastly improve/simplify the development process by allowing developers to safely deploy new features to production. They also allow us to quickly disable features that are causing problems in production.
+
+They allow us to hide features that are still in development/still being tested. This enables a developer to add a feature incrementally to the codebase, and only enable it when it is ready for production. Should the feature break, they also allow us to easily disable it without having to roll back all the changes. This is especially useful when the feature is large and/or complex.
+
+## How do you, as a developer, use feature flags?
+
+### When to use feature flags?
+
+Feature flags should be used when you are working on a feature whose scale makes it hard to implement it all at once and will need to be added incrementally/if the feature is likely to cause breakages. Essentially, feature flags are a must for featues that are not ready for production/fully-tested yet.
```
seanlip: ...whose scale makes it hard to implement all at once, or for more experimental features that are likely to cause breakages. Essentially, ...

Developing-new-features-with-feature-gating.md
```diff
@@ -93,9 +132,37 @@ We use the following principles to determine the value of a feature flags:
...
+
+### How to use feature flags?
+
+Say you are working on a large scale user-facing feature that will take multiple PRs to be fully implemented. The following is the recommended way to use feature flags in such a scenario (*recommended implies that this is the way we use/have used feature flags in the past in Oppia. We haven't encountered a scenario where we required an alternate workflow. Unless there are compelling reasons, this is the development workflow we expect developers to follow when it comes to employing the use of feature flags*):
```
seanlip: to be fully implemented --> to fully implement

I don't think you need all the caveats -- just say that "this is the way". We'll update the wiki in the future if it turns out to not be the way :)

Developing-new-features-with-feature-gating.md
```diff
@@ -93,9 +132,37 @@ We use the following principles to determine the value of a feature flags:
...
+
+1. The very first PR you make must include changes that add a new feature flag to the codebase, meant to be used with this new feature. The feature flag must be placed in the DEV stage. Every single user-facing aspect of the feature -- whether frontend or backend -- must be gated behind the feature flag (both in the first PR, and all the following ones as well). This is to ensure that the feature is not visible to the user until it is fully implemented and ready for production.
+
+2. The first PR must be merged before any following PRs are merged. This is to ensure that the feature flag is available in the codebase for it to be used in the following PRs.
```
seanlip: The first PR --> This first PR above

Developing-new-features-with-feature-gating.md
seanlip: I think it's fine to keep it in TEST, yeah.

Developing-new-features-with-feature-gating.md
```diff
@@ -93,9 +132,37 @@ We use the following principles to determine the value of a feature flags:
...
+
+3. The very last PR you make (to finish up the feature you are working on) must include changes that move the feature flag to the TEST stage. This is to ensure that the feature is available in the test environment, and we can feature-test it before it is made available to the users in the production environment.
```
seanlip: Add to the end: Please test manually to make sure that the feature works fully end-to-end on your local dev server before flipping the flag to TEST.

Developing-new-features-with-feature-gating.md
```diff
@@ -93,9 +132,37 @@ We use the following principles to determine the value of a feature flags:
...
+
+4. If the feature testing reveals that the feature is not ready for production, you must work on fixing the highlighted issues before proceeding further.
 
-For example, if we want to make a feature flag to be enable only in dev environments, we can set the feature flag as the image below.
+5. Once you receive a go ahead from the testers, you must merge another PR, moving the feature flag to the PROD stage, allowing it to be enabled/disabled in production (by the admin(s)).
```
seanlip: go ahead --> go-ahead

testers --> feature testers

you must merge another PR, moving ... --> you must merge another PR. This PR should only do one thing, which is to change the feature flag from the TEST state to the PROD state, ...

Add at the end: When submitting this PR, include a link to the testing doc or other proof that the feature has been approved for release.

Developing-new-features-with-feature-gating.md
```diff
@@ -93,9 +132,37 @@ We use the following principles to determine the value of a feature flags:
...
 
-![Example feature setting](https://i.imgur.com/GnFQ7El.jpg)
\ No newline at end of file
+6. Once the feature is confirmed to be functioning as intended, you must merge one last PR to essentially "un-gate" the feature. Note that the flag itself will remain in the codebase (i.e. in the platform_parameter_list.py file, etc.), it just won't be used anymore (for example, all the `if` blocks and such that were used to gate the feature until this point must be removed).
```
seanlip: as intended --> as intended in production for at least 2 weeks

Shouldn't we also just get rid of the flag altogether at this point? I don't see why we need to keep it in the parameter list.

seanlip (changes requested):
Thanks @nithinrdy, this looks great! Took a pass, PTAL.

Developing-new-features-with-feature-gating.md
nithinrdy: The unit tests aren't about the flags being on/off. Basically, there are two specific tests (one in the backend and one in the frontend) whose success or failure depends on them being aware of all the feature flags that currently exist.

So when a new feature flag is registered in the codebase by a developer, they need to make these unit tests aware of that, otherwise these two tests will fail. Hence it is a part of the feature-flag creation process.

Does that answer the question?

Developing-new-features-with-feature-gating.md
nithinrdy: it: The flag
default state: when the flag is first added to the codebase.

Modified the doc to include these details.

Developing-new-features-with-feature-gating.md
nithinrdy: Yep, modified this point

Developing-new-features-with-feature-gating.md
nithinrdy: Done.

Developing-new-features-with-feature-gating.md
nithinrdy: > Shouldn't we also just get rid of the flag altogether at this point? I don't see why we need to keep it in the parameter list.

I think the flag, once added to the codebase and after making its way into the prod server, ends up in the datastore. Since flag names must be unique, the flag is moved to the deprecated list (I've updated the instructions) to allow for developers to see how to name new feature flags so there aren't any conflicts.

I have no source for this information (the feature flags feature isn't documented all that well), this is just my guess as to why we can't just remove the feature flags and instead have to move them into the deprecated list and have them stick around even after we're done using them.

nithinrdy:
@seanlip PTAL

Developing-new-features-with-feature-gating.md
seanlip: Yup (though maybe you could explain this a bit more in the description of these changes).

But, in addition to the above, do you want to explain to developers how to test stuff that is gated by a feature flag? This should probably be part of "standard procedure", too.

Developing-new-features-with-feature-gating.md
```diff
@@ -149,20 +149,20 @@ They allow us to hide features that are still in development/still being tested.
...
+Feature flags should be used when you are working on a feature whose scale makes it hard to implement all at once, or for more experimental features that are likely to cause breakages. Essentially, feature flags are a must for features that are not ready for production/fully-tested yet.
 
 ### How to use feature flags?
 
-Say you are working on a large scale user-facing feature that will take multiple PRs to be fully implemented. The following is the recommended way to use feature flags in such a scenario (*recommended implies that this is the way we use/have used feature flags in the past in Oppia. We haven't encountered a scenario where we required an alternate workflow. Unless there are compelling reasons, this is the development workflow we expect developers to follow when it comes to employing the use of feature flags*):
+Say you are working on a large scale user-facing feature that will take multiple PRs to fully implement. The following is the recommended way to use feature flags in such a scenario (*this is the development workflow we expect developers to follow when it comes to employing the use of feature flags*):
```
seanlip: I'd go even stronger: the second sentence could just be "In such cases, please use feature flags to gate your feature appropriately by carefully following the steps below:"

Developing-new-features-with-feature-gating.md
seanlip: Ah, I see. Yeah I think you're right about them moving into the datastore. In step 6 can you also ask developers to also remove the flag entirely from the codebase except from that one part?

seanlip:
Thanks @nithinrdy! Left additional replies; see also https://github.com/oppia/oppia-web-developer-docs/pull/86#discussion_r1088724714

nithinrdy:
Got it, working on them.

Developing-new-features-with-feature-gating.md
nithinrdy: I've added instructions for unit testing gated features in the frontend.

However, I'm not very well-versed in Oppia's backend, nor do I have much experience writing unit tests in it. Have any of the other contributors worked with the gating system on the backend? Meanwhile, I'll try looking into `platform_feature_services_test.py` to try and figure out how flags are tested there.

nithinrdy:
Working on adding instructions for unit testing gated features in the backend. PTAL at the rest!

Developing-new-features-with-feature-gating.md
seanlip: This I don't really know, sorry -- though @BenHenning might since he originally oversaw the project. I think looking at the file you mentioned is a good first step, FWIW.

@BenHenning if you have any comments on this, could you please leave them here? Thanks!

seanlip (commented):
Hi @nithinrdy, thanks -- this really looks fantastic, and it's just the one comment that's left. I don't have anything else to add! Thanks a lot for writing this up :)

Developing-new-features-with-feature-gating.md
```diff
@@ -93,9 +199,37 @@ We use the following principles to determine the value of a feature flags:
...
-- Each filter describes a scenario, there are multiple types of filters like `server mode`, `client type`, so that you can you can describe scenarios like `server mode = dev OR server mode = test`.
-- Each filter can have multiple conditions, and a filter is matched if any of the conditions are matched.
+- Each filter describes a scenario. There are multiple types of filters (like `server mode` and `client type`) which can be used in combination with each other to describe various scenarios (like `server mode = dev OR server mode = test`).
+- Each filter can have multiple conditions, and a filter is matched if any of the conditions specified in the rule are matched.
+
+For example, if we want a feature flag to be enabled only in dev environments, we can configure the flag's rules/filters as shown below.
+
+![Example feature setting](images/adminPageFeatureFlagSettings.png)
+
+## Why do we need feature flags?
```
BenHenning, line 294: Related to what's here, but another benefit is it lets us decouple the feature & binary releases which allows us to release the server with less risk of regressions.

Developing-new-features-with-feature-gating.md
```diff
@@ -93,9 +199,37 @@ We use the following principles to determine the value of a feature flags:
...
+
+Feature flags vastly improve/simplify the development process by allowing developers to safely deploy new features to production. They also allow us to quickly disable features that are causing problems in production.
+
+They allow us to hide features that are still in development/still being tested. This enables a developer to add a feature incrementally to the codebase, and only enable it when it is ready for production. Should the feature break, they also allow us to easily disable it without having to roll back all the changes. This is especially useful when the feature is large and/or complex.
+
+## How do you, as a developer, use feature flags?
+
+### When to use feature flags?
+
+Feature flags should be used when you are working on a feature whose scale makes it hard to implement all at once, or for more experimental features that are likely to cause breakages. Essentially, feature flags are a must for features that are not ready for production/fully-tested yet.
```
BenHenning:
```suggestion
Feature flags should be used when you are working on a feature whose scale makes it hard to implement all at once, or for more experimental features that are likely to cause breakages. Essentially, feature flags are a must for features that are not ready for production/fully-tested at the time that they're merged into develop.
```

Bit clearer on what 'yet' means. @seanlip do you agree with this framing?

Developing-new-features-with-feature-gating.md
```diff
@@ -93,9 +199,37 @@ We use the following principles to determine the value of a feature flags:
...
 
-For example, if we want to make a feature flag to be enable only in dev environments, we can set the feature flag as the image below.
+5. Once you receive a go-ahead from the feature testers, you must merge another PR -- this PR is meant to do only one thing, i.e. move the feature flag to the PROD stage, allowing it to be enabled/disabled in production (by the admin(s)). **NOTE: When opening this PR, include a link to the testing doc or other proof that the feature has been approved for release.**
 
-![Example feature setting](https://i.imgur.com/GnFQ7El.jpg)
\ No newline at end of file
+6. Once the feature is confirmed to be functioning as intended in production (for at least two weeks), you must merge one last PR to essentially "un-gate" the feature and move the feature flag to the deprecated stage (one of the stages listed in `core/domain/platform_parameter_list.py` for flags that are no longer in use). Then feel free to remove all references of the feature flag from the codebase (for example, in all the `if` blocks you created to gate the feature).
```
BenHenning: 'Two weeks' feels a bit arbitrary. Shouldn't it be after landing criteria are met and no regressions are detected via stackdriver and other usage metrics? This may vary from feature to feature. @seanlip WDYT?

Developing-new-features-with-feature-gating.md
```diff
@@ -93,9 +199,37 @@ We use the following principles to determine the value of a feature flags:
 
 - Each feature flag has multiple *rules*, its value changes to the value specified in the first rule that is matched. If no rule is matched, its value remains default.
 - Each rule has multiple *filters*, and a rule is matched only when all of its filters are matched.
-- Each filter describes a scenario, there are multiple types of filters like `server mode`, `client type`, so that you can you can describe scenarios like `server mode = dev OR server mode = test`.
-- Each filter can have multiple conditions, and a filter is matched if any of the conditions are matched.
+- Each filter describes a scenario. There are multiple types of filters (like `server mode` and `client type`) which can be used in combination with each other to describe various scenarios (like `server mode = dev OR server mode = test`).
+- Each filter can have multiple conditions, and a filter is matched if any of the conditions specified in the rule are matched.
+
+For example, if we want a feature flag to be enabled only in dev environments, we can configure the flag's rules/filters as shown below.
```
BenHenning: FWIW this example may be a bit confusing as it illustrates yet another way to gate on environment type. Perhaps giving an example of blocking for a specific browser might be a bit more clear?

BenHenning (changes requested):
Thanks @nithinrdy. Left some thoughts--PTAL!

Developing-new-features-with-feature-gating.md
seanlip: I think Ben's suggestion is fine -- would just suggest changing "not ready" to "not yet ready".

Developing-new-features-with-feature-gating.md
seanlip: Actually I think two weeks is OK for most things. The idea is to have it running in prod for at least two weeks without issues which I think is reasonable, and I think there should be a rough timeframe here to set expectations.

But to your point, we can include a definition of what "functioning as intended" means. Perhaps something like this:

```
Once the feature is confirmed to be functioning as intended in production (for at least two weeks), please do the following:

- Make sure that the feature is really functioning as intended, and ready to be made permanent. To do this, confirm with the PMs that no users have reported issues with it, and that no regressions have been detected via StackDriver and other usage metrics.
- Once you have confirmation that the feature can be made permanent, merge one last PR to "un-gate" the feature and move the feature flag to the deprecated stage (one of the stages listed in `core/domain/platform_parameter_list.py` for flags that are no longer in use). In that PR, please also remove all remaining references to the feature flag from the codebase (for example, in all the `if` blocks you created to gate the feature).
```

@nithinrdy does this sound good to you?

Developing-new-features-with-feature-gating.md
BenHenning: I would've suggested the same thing (to look at existing cases of platform parameters being used in tests both in Oppia web and backend). I'm not sure there's better references or documentation available right now other than the actual code, however.

Regarding the actual best practice here, I think we should encourage testing functionality both with the flag on and off (since both states will exist for users at some point in time, so we need to make sure that the correct thing happens in both case rather than relying on the default flag state).

Developing-new-features-with-feature-gating.md
seanlip: @nithinrdy Could you instead show an example of "this is an example of a flag that works in test + prod mode, but not in dev mode"? It's not intuitive that we need to add one rule per server (especially since the dev/test/prod FeatureStage criteria behave differently), so I think it is worth calling out explicitly that every environment needs to be explicitly mentioned in the admin panel settings.

Developing-new-features-with-feature-gating.md
```diff
@@ -93,9 +199,37 @@ We use the following principles to determine the value of a feature flags:
...
+
+2. The first PR above must be merged before any following PRs are merged. This is to ensure that the feature flag is available in the codebase for it to be used in the following PRs.
+
+3. The very last PR you make (to finish up the feature you are working on) must include changes that move the feature flag to the TEST stage. This is to ensure that the feature is available in the test environment, and we can feature-test it before it is made available to the users in the production environment. **NOTE: Please test all the changes manually to make sure that the feature works fully end-to-end on your local dev server, before flipping the flag to TEST.**
+
+4. If the feature testing reveals that the feature is not ready for production, you must work on fixing the highlighted issues before proceeding further.
```
seanlip: Maybe add, at the start of step 4, something like this: "In the test server's /admin page, turn the feature flag on for "test" mode and ask the product/QA teams to help with feature testing your feature."

(TBH I think this process is a bit silly ... if a feature flag is in the test featurestage then we should just treat it as enabled for dev/test, and if it's in the prod featurestage then we should treat it as enabled for dev/test/prod. The setting on the /admin page should then be for whether to enable or disable it for the server it is currently running on, and shouldn't be conditional on what the server mode is -- since that's already known for the server it's running on. But we can fix this later; let's just document the current procedure.)

Developing-new-features-with-feature-gating.md
```diff
@@ -93,9 +199,37 @@ We use the following principles to determine the value of a feature flags:
...
 
-For example, if we want to make a feature flag to be enable only in dev environments, we can set the feature flag as the image below.
+5. Once you receive a go-ahead from the feature testers, you must merge another PR -- this PR is meant to do only one thing, i.e. move the feature flag to the PROD stage, allowing it to be enabled/disabled in production (by the admin(s)). **NOTE: When opening this PR, include a link to the testing doc or other proof that the feature has been approved for release.**
```
seanlip: Add after this: Once this PR is merged, send a request to the release coordinators to turn on the feature in production by adding a rule in the /admin page.

Developing-new-features-with-feature-gating.md
nithinrdy: Done.

Developing-new-features-with-feature-gating.md
nithinrdy: Yep, rewrote instruction 6 with these steps.

Developing-new-features-with-feature-gating.md
nithinrdy: You're right, changed it to rules on browser types.

nithinrdy: Done (except example shows rules for enabling feature on Chrome + Firefox. Same key point as you brought up, just with browsers).

Developing-new-features-with-feature-gating.md
nithinrdy: > add, at the start of step 4, something like this: "In the test server's /admin page, turn the feature flag on for "test" mode and ask the product/QA teams to help with feature testing your feature."

Do general contributors have access to the test server? I think enabling the feature via the admin page for testing is something the testers can take care of, followed by any testing they need to do. (I could add something like "Note to testers, do this before you start testing...")

Developing-new-features-with-feature-gating.md
nithinrdy: > we should encourage testing functionality both with the flag on and off (since both states will exist for users at some point in time, so we need to make sure that the correct thing happens in both case rather than relying on the default flag state)

Yep, that's actually how the frontend tests are currently written (mock `PlatformFeatureService` and test with the feature status getter set to return both `false` and `true` in two separate tests). Makes sense to test the backend side of things in a similar manner.

nithinrdy:
Addressed all comments!

Also, @seanlip [PTAL](https://github.com/oppia/oppia-web-developer-docs/pull/86#discussion_r1104078556).

The backend tests are all that's left, I'm looking into them.

Although Sean, since you mentioned how contributors are currently a little stuck on the gating process -- for now, simply gating the `xyz-backend-api.service` in the frontend (more specifically, the methods in these services that make the HTTP requests) is a good enough alternative to directly gating the backend controllers. What do you think?

Developing-new-features-with-feature-gating.md
seanlip: Can we ask the contributors to ask the release admins to turn it on? That would probably be best, because there needs to be some sort of alert to the release testing admins that a feature needs to be enabled.

seanlip:
Thanks @nithinrdy! Looks good, left one comment here: https://github.com/oppia/oppia-web-developer-docs/pull/86#discussion_r1104130615

Hm, I think that we should gate the backend controllers at least, because it'll be possible to make direct requests via wget/curl to them otherwise. When I said that folks are stuck on the gating process, I meant that there's general confusion about how feature flags work and that's causing e.g. features not to be enabled or tested on the test server despite being planned for the next release. So the fix for that is publishing this wiki page so that everyone has a single source of truth to refer to :)

Developing-new-features-with-feature-gating.md
```diff
@@ -135,34 +202,40 @@ We use the following principles to determine the value of a feature flags:
...
+
+6. Once this PR is merged, send a request to the release coordinators to turn on the feature in production by adding a rule in the `/admin` page.
 
-6. Once the feature is confirmed to be functioning as intended, you must merge one last PR to essentially "un-gate" the feature. Note that the flag itself will remain in the codebase (i.e. in the platform_parameter_list.py file, etc.), it just won't be used anymore (for example, all the `if` blocks and such that were used to gate the feature until this point must be removed).
+7. Once the feature is confirmed to be functioning as intended in production (for at least two weeks), please do the following, in order:
+    - Make sure that the feature is indeed functioning as intended, and ready to be made permanent. To do this, confirm with the PMs that no users have reported issues with it, and that no regressions have been detected via StackDriver and other usage metrics.
```
BenHenning: Can we make the "other usage metrics" more concrete? I suspect that people won't actually know what to do for this part.

Developing-new-features-with-feature-gating.md
BenHenning: @seanlip I'm not sure that I understand your reasoning--it actually seems equally arbitrary to the original phrasing. Given that some features might be less discoverable or not on the main happy path, is 2 weeks *actually* enough time to be confident in the feature not causing regressions?

BenHenning (commented):
Thanks--left a few follow-ups.

Developing-new-features-with-feature-gating.md
seanlip: @nithinrdy Let's say "via StackDriver or general user feedback".

Developing-new-features-with-feature-gating.md
seanlip: To be clear, I'm thinking of 2 weeks as the minimum, not the maximum. This doesn't guarantee that 2 weeks is enough time. It's saying that any feature should be in prod for at least that long. The _actual_ criteria is that "the feature works correctly".

So in the scenario you describe, I imagine that the developer might ask the PM after 2 weeks "are we confident?" and the PM would say "not yet, wait a bit longer". That's WAI.

Do you see an issue with this approach?

Developing-new-features-with-feature-gating.md
nithinrdy: Got it.

Developing-new-features-with-feature-gating.md
nithinrdy: Yep, makes sense.

Developing-new-features-with-feature-gating.md
nithinrdy: Done.

nithinrdy:
@seanlip @BenHenning I've added the instructions for writing backend tests; PTAL (I tested them using an existing feature flag and gating a controller with it -- they're working as expected).

Developing-new-features-with-feature-gating.md
```diff
@@ -145,6 +145,91 @@ spyOnProperty(platformFeatureService, 'status', 'get').and.returnValue(
...
+```python
+from core.domain import platform_feature_services as feature_services # (alias is optional)
+from core.domain import platform_parameter_domain
+from core.domain import platform_parameter_list
+from core.domain import platform_parameter_registry as registry # (alias is optional)
+```
+
+2. Clear the parameter registry and create a mock feature flag (using the actual enum of the feature flag -- we recommend you do it in the `setUp` hook), like so:
+
+```python
```
seanlip, line 163: @nithinrdy For consistency with the codebase, could you please use 4-space indentation in all Python code blocks?

Developing-new-features-with-feature-gating.md
```diff
@@ -145,6 +145,91 @@ spyOnProperty(platformFeatureService, 'status', 'get').and.returnValue(
...
+self.mock_feature_flag = registry.Registry.create_feature_flag(
+  platform_parameter_list.ParamNames.NewFeature, 'a mock of the new_feature feature flag',
+  platform_parameter_domain.FeatureStages.DEV) # FeatureStage DEV just simplifies things
```
seanlip: Can you explain more what you mean by "FeatureStage DEV just simplifies things"? I don't really understand what that's trying to convey.

Developing-new-features-with-feature-gating.md
```diff
@@ -145,6 +145,91 @@ spyOnProperty(platformFeatureService, 'status', 'get').and.returnValue(
...
+                        }
+                    ],
+                    'value_when_matched': True
+                })
+            ]
+        )
+
+```
+
+**NOTE**: Like mentioned above `self.owner_id` refers to the id of the user with the necessary permissions to perform the actual feature flag rules update. Some tests have the `owner_id` property created within the `setUp()` hook, and this variable can be used throughout the test class.
```
seanlip: Can you explain what permissions owner_id must have and how to make sure they have them? (They will need superadmin permissions, right?)

Developing-new-features-with-feature-gating.md
```diff
@@ -230,12 +315,14 @@ Say you are working on a large scale user-facing feature that will take multiple
 
 3. The very last PR you make (to finish up the feature you are working on) must include changes that move the feature flag to the TEST stage. This is to ensure that the feature is available in the test environment, and we can feature-test it before it is made available to the users in the production environment. **NOTE: Please test all the changes manually to make sure that the feature works fully end-to-end on your local dev server, before flipping the flag to TEST.**
 
-4. If the feature testing reveals that the feature is not ready for production, you must work on fixing the highlighted issues before proceeding further.
+4. Ask the release testers to enable the feature flag of the feature you are building for the TEST environment (in the test server, of course), before proceeding with the actual testing.
```
seanlip: Don't need the "of course" :) Also the release testers in general won't be able to manipulate feature flags. I suggest:

"Ask the release coordinator to go to the test server admin page and enable the feature flag for the feature you are building, before opening up the server for the actual testing."

seanlip (changes requested):
Thanks, @nithinrdy! Took a pass, just a few comments (but they should be straightforward, I think). PTAL.

Developing-new-features-with-feature-gating.md
nithinrdy: Done.

Developing-new-features-with-feature-gating.md
nithinrdy: My bad, I meant to imply that I chose (and so should the developer reading it) the DEV stage for the sake of this example (or for the sake of simplicity). I just phrased it weirdly.

Developing-new-features-with-feature-gating.md
nithinrdy: Added that the code block shows how to access the owner account credentials, and that they provide super admin permissions to an account created using them, which are used to update the feature flag rules.

> how to make sure they have them

An account made using the `self.OWNER_EMAIL, self.OWNER_USERNAME` credentials, which are a part of the `GenericTestBase` class and are available to every test class, has those permissions. To make sure the created account has these permissions, the developer needs to use those credentials to create the account.

Developing-new-features-with-feature-gating.md
nithinrdy: Done!

nithinrdy:
@seanlip PTAL.

seanlip (approved):
Thanks @nithinrdy. LGTM, no further comments from my end!

@BenHenning would you please take a pass so that this can be merged? Thanks!

Developing-new-features-with-feature-gating.md
BenHenning: I think if that's the intent, then it should be codified as such. Based on that, I suggest rewording to:

``... confirmed with the product team to be functioning as intended in production (for no fewer than two weeks)``

Or something similar--WDYT @seanlip @nithinrdy?

BenHenning (commented):
Thanks! Just had one follow-up, otherwise the PR LGTM.

Developing-new-features-with-feature-gating.md
seanlip: Hi @BenHenning ... I think that's fine, though I'm not really sure how "at least 2 weeks" is different from "no fewer than 2 weeks". Aren't they logically/mathematically the same? I'm happy with either wording though, if this one is preferred.

BenHenning: They're equivalent, but for some reason I read "at least 2 weeks" as "test for 2 weeks" rather than "test for 2 weeks **minimum**". The verbiage change here is meant to make the point a bit stronger, though if this is a me-only interpretation feel free to ignore that suggestion. The "confirmed with the product team" is the more important addition.

nithinrdy: Sounds good! They mean almost the same thing, although I suppose it's good to be as verbose as possible since the idea is to have the contributors follow this workflow to a tee; the clearer the instructions, the better a contributor can stick to them.

nithinrdy:
@seanlip @BenHenning PTAL. I think the PR's good to go; Please merge it if you see no other issues with it.

seanlip (approved):
Great, thanks @nithinrdy! LGTM.

@BenHenning could you PTAL and approve if it looks good?  Thanks!

BenHenning (approved):
Thanks @nithinrdy. This LGTM!

PR #88, opened 2023-01-17
Updates installation instructions for Linux

Patel-Muhammad:
This pull request updates the instructions for setting up the **`pyenv`** environment and makes it simpler to add the necessary lines to the **`.bashrc`** file by utilizing the echo command.

@U8NWXD, would you please review this pull request?

### DIFF
--- a/Installing-Oppia-(Linux;-Python-3).md
+++ b/Installing-Oppia-(Linux;-Python-3).md
@@ -126,16 +126,17 @@ For your virtual environment, we recommend you use [pyenv](https://github.com/py
    ...
    ```
 
-   If you see the warning at the end, add the following lines to your `.bashrc` (see [here](https://askubuntu.com/a/127059) for where to find this file):
-
+     If you see a warning message, please run the following commands to fix it. These commands will add necessary lines to a file called `.bashrc` which is located in your home directory. (The `.bashrc` file is used to set some environment variable. To know more, please refer to this [link](https://askubuntu.com/a/127059)):
+   
    ```bash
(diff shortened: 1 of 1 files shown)


Patel-Muhammad:
@seanlip can you please review this PR?

seanlip:
This is a dev workflow change so I've assigned the dev workflow leads.

Personally I don't think this PR is needed / helps things, but let's see what they say.

U8NWXD:
I agree with @seanlip. We expect contributors to be comfortable editing text files (they are writing code after all), so I think it's totally reasonable to ask them to add some lines to `~/.bashrc`.

Further, I think using `echo` actually makes the instructions more difficult to follow because readers can no longer copy the text that needs to be added to `~/.bashrc`, and it's harder to understand what the lines being added mean (e.g. we lose the bash syntax highlighting).

U8NWXD:
I'm open to being convinced otherwise if you still disagree @Patel-Muhammad, but right now I don't think this is a change we want to make

Patel-Muhammad:
Thank you for the update and for assigning the dev workflow leads. Looking forward to hear the feedback from the leads.

U8NWXD:
Sorry, I should have made clear that I'm one of the dev workflow leads :)

Patel-Muhammad:
@U8NWXD  I understand. Thank you for letting me know about your perspective, as a new contributor it will help my own perspective to get aligned with yours.

Patel-Muhammad:
So, now should I close the pull request?

seanlip:
Yes, since it's not needed. I'll do that.

PR #91, opened 2023-01-18
Fix #52: Update revert and regression policy

### LINKED ISSUE #52: Clarify revert and regression policy in the case where new PRs are merged on top of the "problematic" PR
Per discussion with @heyimShivam, it would be good if this page:

https://github.com/oppia/oppia/wiki/Revert-and-Regression-Policy 

had additional clarification on the procedure for fixing regressions in the explicit case where new PRs are merged on top of the PR that caused the regression.

------

(My take is that we should still revert the bad PR, and also try to enforce small PRs that are easily revertable in general. But open to other thoughts here, before we write this down as the rule. I'm also open to anything that keeps develop clean -- maybe we can say that the developer who introduced the issue should be contacted and must fix it within 24 hours, otherwise the revert will happen, and the revert might happen earlier than that at the discretion of the tech lead -- but again, open to suggestions here.)

U8NWXD:
Fixes #52 by updating our policy for reverting PRs to fix regressions


Revert-and-Regression-Policy.md
```diff
@@ -1,18 +1,72 @@
...
+When a PR is merged that causes a regression in develop, there are two ways to fix the regression:
 
-3. If the regression causes problems for developers (e.g. failing tests), send an email to oppia-dev to let developers know. The email should include a link to the issue you opened so that developers can follow along as we fix the regression.
+1. Revert the PR. This can be quick if there are no merge conflicts, but with large PRs, reverts can be difficult.
+2. Fix forward by merging another PR without reverting the problematic PR. This can be easy if the bug is small and easily-identifiable, but if the solution is unclear, fixing forward can be hard.
 
-3. Investigate to find the cause of the regression. Ideally, we want to find the PR that introduced the regression, but failing that, we want to understand what problem in the code is causing the regression.
+Note that in either case, the PR author is responsible for fixing the bug in their PR, either by re-creating their PR without the bug (if we revert) or by opening a new PR to fix the bug (if we fix forward).
```
gp201:
```suggestion
> **Note**
Note that in either case, the PR author is responsible for fixing the bug in their PR, either by re-creating their PR without the bug (if we revert) or by opening a new PR to fix the bug (if we fix forward).
```

Revert-and-Regression-Policy.md
```diff
@@ -23,72 +77,37 @@
...
+
 3.  Run `git bisect bad` to verify that the current commit contains the bug.
-    
+
 4.  Go to [this](https://github.com/oppia/oppia/commits/develop) page to find an older (“good”) commit where the bug did not exist. Copy the commit id of the commit, and do git checkout on it (git checkout <commit-id>). Start the dev server (by running the start script) and try to figure out whether the error still exists. If so, you will need to find an even older commit than the current one. Kill the server, and repeat this step until you find a “good” commit.
-    
+
 5.  Once you’ve found a “good” commit, run “git bisect good”. The tool will automatically perform a binary search and checkout a commit between the “good” commit and “bad” commit. Start the dev server (by running the start script) and try to reproduce the error. At each step:
     - If the bug is reproducible, run “git bisect good”.
     - If not, run “git bisect bad”.
```
gp201:
```suggestion

5.  Once you’ve found a “good” commit, run “git bisect good”. The tool will automatically perform a binary search and checkout a commit between the “good” commit and “bad” commit. Start the dev server (by running the start script) and try to reproduce the error. At each step:
    - If the bug is reproducible, run `git bisect good`.
    - If not, run `git bisect bad`.
```

Revert-and-Regression-Policy.md
```diff
@@ -23,72 +77,37 @@
...
 If you see this:
 
-![](https://lh3.googleusercontent.com/4uYWOO2yXW8eBVTG9BH-wYZKRo7rB2WNUBbBudtaprlp4btYJ3avdQP-fRnZSBXGAq1DM05Hc_c9haASvHTFF8gRHgPZNqQ3ZKZc7vkZrPy13rTZ2EwOuQXQM6Sz7j0gVpF-61Z5)  
-  
-You will need to make the code changes on your local machine and create a PR by hand.
-
-## What do I do if I caused a regression?
-
-If the PR that introduced the regression hasn't been reverted, please follow the steps above to revert. Once your PR is reverted, follow these steps:
+![](https://lh3.googleusercontent.com/4uYWOO2yXW8eBVTG9BH-wYZKRo7rB2WNUBbBudtaprlp4btYJ3avdQP-fRnZSBXGAq1DM05Hc_c9haASvHTFF8gRHgPZNqQ3ZKZc7vkZrPy13rTZ2EwOuQXQM6Sz7j0gVpF-61Z5)
```
gp201: Where is this image stored?

gp201 (changes requested):

Revert-and-Regression-Policy.md
U8NWXD: Someone's Google account. I've downloaded the images and checked them into git

Revert-and-Regression-Policy.md
U8NWXD: done

Revert-and-Regression-Policy.md
U8NWXD: done

gp201 (approved):
LGTM

PR #98, opened 2023-01-24
Removed  Debian reference from Linux install guide

prajapati-arjun:
Removed Debian reference from Linux install guide.


seanlip (changes requested):
@prajapati-arjun What issue does this fix and why are we doing it? Please link to the relevant issue or GitHub Discussion.

prajapati-arjun:
issue : https://github.com/oppia/oppia/issues/17113

seanlip:
Ah, please make sure to always read the whole comment thread before making a PR (or posting). We decided to go with a different solution. I updated the title of the issue, so this PR is probably now invalid.

PR #102, opened 2023-01-27
Make it clearer how to choose an issue. Other minor changes.

seanlip:
This PR does a few things:

- Clarifies the process for choosing a good first issue
- Points to key instructions from the GSoC page and asks contributors to read those carefully
- Adds some notes to the PR process on adding a good description for the PR

### DIFF
--- a/Contributing-code-to-Oppia.md
+++ b/Contributing-code-to-Oppia.md
@@ -65,7 +65,7 @@ Welcome! Please make sure to follow the instructions above if you haven't alread
 
 After that, you can choose a good first issue from the [list of good first issues](https://github.com/oppia/oppia/labels/good%20first%20issue). These issues are hand-picked to ensure that you don't run into unexpected roadblocks while working on them, and each of them should have clear instructions for new contributors. If you see one that doesn't, please let us know via [GitHub Discussions](https://github.com/oppia/oppia/discussions) and we'll fix it. For other issues, you might need to be more independent because we might not know how to solve them either.
 
-You can also browse good first issues for each of the core Oppia Web teams to find something you'd enjoy working on. Please only choose issues that have **not yet** been assigned! Here are the project boards for the different teams:
+You can also browse good first issues for each of the core Oppia Web teams to find something you'd enjoy working on! Please only choose issues that have **not yet** been assigned, unless the issue is a "checkbox issue" with multiple claimable parts. Here are the project boards for the different teams:
 
 - Learner and Creator Experience (LaCE): https://github.com/orgs/oppia/projects/3/views/10, typically frontend or full-stack
 - Contributor Dashboard: https://github.com/orgs/oppia/projects/18/views/5, typically frontend or full-stack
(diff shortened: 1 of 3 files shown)


seanlip:
@U8NWXD @gp201 PTAL. Thanks!

Contributing-code-to-Oppia.md
```diff
@@ -65,21 +65,30 @@ Welcome! Please make sure to follow the instructions above if you haven't alread
 
 After that, you can choose a good first issue from the [list of good first issues](https://github.com/oppia/oppia/labels/good%20first%20issue). These issues are hand-picked to ensure that you don't run into unexpected roadblocks while working on them, and each of them should have clear instructions for new contributors. If you see one that doesn't, please let us know via [GitHub Discussions](https://github.com/oppia/oppia/discussions) and we'll fix it. For other issues, you might need to be more independent because we might not know how to solve them either.
 
-You can also browse good first issues for each of the core Oppia Web teams to find something you'd enjoy working on:
+You can also browse good first issues for each of the core Oppia Web teams to find something you'd enjoy working on. Please only choose issues that have **not yet** been assigned! Here are the project boards for the different teams:
```
gp201: What about checkbox issues? They're assigned to the dev who monitors and assigns the checkboxes. This may mislead contributors to think they can't work on them.

gp201 (changes requested):
Thanks, @seanlip ! I left some comments. PTAL

Contributing-code-to-Oppia.md
seanlip: Good catch! Done.

seanlip:
@gp201 PTAL. (Also please reassign me next time if you want me to do something, per our standard PR procedure.)

@U8NWXD PTAL too if you'd like to. Thanks!

Contributing-code-to-Oppia.md
```diff
@@ -65,21 +65,30 @@ Welcome! Please make sure to follow the instructions above if you haven't alread
...
+Once you have a good understanding of the issue, you can ask for it to be assigned to you by leaving a comment as follows:
+
+- Explain clearly how you'd tackle the issue (at a minimum, point to which file(s) you'd modify and describe the changes you'd make).
+- If possible, show a screenshot or code snippet demonstrating your proposed fix.
+- @-mention the leads of the corresponding project the issue falls under, letting them know you'd like to work on it. (The leads are: @SanjaySajuJacob and @Priyansh61 for [LaCE](https://github.com/orgs/oppia/projects/3/views/8?pane=info); @sagangwee and @chris7716 for [Contributor Dashboard](https://github.com/orgs/oppia/projects/18/views/4?pane=info); @U8NWXD and @gp201 for [Developer Workflow](https://github.com/orgs/oppia/projects/8?pane=info).)
```
U8NWXD, line 87: I worry that this will become out of date as leads change. How about we make sure leads are specified on the project boards and point people to that? This way there will be a single source of truth for who the leads are

seanlip: @U8NWXD That's what I did before -- it doesn't seem to have worked though, people keep @-mentioning me instead :) So I would suggest, let's try this and see if it works? If so then we can figure out how to make it more long-term-proof at a later stage.

U8NWXD: Sure. My other suggestion would be to indicate the person to contact in the issue, that way contributors don't have to look elsewhere. That's probably going to become out of date even faster though, so your approach might be a good compromise

gp201 (approved):
LGTM

Contributing-code-to-Oppia.md
seanlip: Yeah, agreed, and it'll be a lot of overhead to include on a per-issue basis. So yup, I'd suggest leaving it as-is for now and seeing how it goes.

PR #107, opened 2023-02-07
Clarify revert and regression policy

U8NWXD:
Clarify who can revert PRs and that reverts are blame-free. These issues were raised in https://github.com/oppia/oppia/pull/17195#issuecomment-1416702617

### DIFF
--- a/Revert-and-Regression-Policy.md
+++ b/Revert-and-Regression-Policy.md
@@ -15,7 +15,7 @@
 
 Sometimes, PRs need to be reverted despite the PR author doing nothing wrong. For example, sometimes a race condition occurs where 2 PRs are merged in quick succession and conflict. We may have to revert one of the PRs to fix develop even though no PR author did anything wrong.
 
-It is important to note that reverts are not ment to be punitive. A key tenet of the dev workflow team is that develop needs to be kept correct at all times. When a regression appears, we need to fix develop as quickly as possible to avoid blocking other developers, and sometimes the quickest way is to revert PRs. Thus, we revert PRs to help developers keep working, not because the PR author did something wrong.
+It is important to note that reverts are not meant to be punitive. A key tenet of the dev workflow team is that develop needs to be kept correct at all times. When a regression appears, we need to fix develop as quickly as possible to avoid blocking other developers, and sometimes the quickest way is to revert PRs. Thus, we revert PRs to help developers keep working, not because the PR author did something wrong.
 
 ## Policy
 
--- a/Revert-and-Regression-Policy.md
+++ b/Revert-and-Regression-Policy.md
@@ -43,6 +43,15 @@ To decide whether an exception is available, first decide whether anyone besides
 | Moderate                         | 1 (if points 1b and 1c are met)    |
 | Minor                            | 1 (if points 1b and 1c are met), 2 |
 
+Severity examples:
+
+* Severe: CI checks are failing so frequently that many PRs need to have their
+  tests rerun.
+* Moderate: CI checks fail sometimes, but infrequently enough that only a few
+  PRs get delayed by them.
+* Minor: CI runs are unaffected, but the developer experience is made slightly
+  less intuitive by needing to add an extra flag when running a script.
+
 ## Instructions
 
 ### Developers Who Notice a Regression
--- a/Revert-and-Regression-Policy.md
+++ b/Revert-and-Regression-Policy.md
@@ -13,9 +13,9 @@
 
 ## Blame-Free Reverts
 
-In the past, reverts have sometimes been seen as a punitive measure . This is no longer the case at Oppia. Often, PRs need to be reverted despite the PR author doing nothing wrong. For example, sometimes a race condition occurs where 2 PRs are merged in quick succession and conflict. We may have to revert one of the PRs to fix develop even though no PR author did anything wrong.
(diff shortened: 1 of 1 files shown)


Revert-and-Regression-Policy.md
```diff
@@ -1,14 +1,22 @@
...
   * [Developers Who Cause a Regression](#developers-who-cause-a-regression)
+  * [Developers with Revert Permissions](#developers-with-revert-permissions)
   * [Dev Workflow Leads](#dev-workflow-leads)
 * [Tips](#tips)
   * [How to identify the bad commit](#how-to-identify-the-bad-commit)
   * [How to revert a PR](#how-to-revert-a-pr)
 
+## Blame-Free Reverts
+
+In the past, reverts have sometimes been seen as a punitive measure . This is no longer the case at Oppia. Often, PRs need to be reverted despite the PR author doing nothing wrong. For example, sometimes a race condition occurs where 2 PRs are merged in quick succession and conflict. We may have to revert one of the PRs to fix develop even though no PR author did anything wrong.
```
seanlip: I'm a bit unsure about the first two sentences. I don't think we've ever intentionally applied reverts punitively, though the perception to the revertee may be that reverts are punitive. Both that policy (of not intentionally applying reverts punitively) and perception may well still continue into the future, so I don't think that something has necessarily changed there. The main thing that's changed is that we now have a wiki page that folks can reference and that puts more structure around the process.

So, perhaps start this paragraph with "Sometimes, PRs need to be reverted because of XXX." and omit the first two sentences, then have the second paragraph start with "It is important to note that reverts are not meant to be punitive. A key tenet ..."?

Revert-and-Regression-Policy.md
```diff
@@ -56,17 +64,20 @@ You should also begin debugging your changes to figure out what caused the regre
 
 Once the dev workflow leads decide whether to revert, fix forward, or both, you should start work on the associated actions (re-creating the PR with the bug fixed, opening a fix-forward PR to fix the bug, or both).
 
+### Developers with Revert Permissions
+
+Any Oppia developer with permission to open revert PRs on GitHub may decide, following the above policy, whether to revert the PR, fix forward, pursue both in parallel, or make one of the two allowed exceptions. They should document their reasoning and communicate on the tracking issue so the dev workflow leads can understand what happened when they consider mitigations.
```
seanlip: What does "when they consider mitigations" mean?

Revert-and-Regression-Policy.md
```diff
@@ -56,17 +64,20 @@ You should also begin debugging your changes to figure out what caused the regre
...
+
 ### Dev Workflow Leads
 
-Once you are informed of a regression in develop, you should make sure a tracking issue is created and alert the rest of the dev workflow leads. Decide, following the above policy, whether to revert the PR, fix forward, pursue both in parallel, or make one of the two allowed exceptions. Once a decision is made, communicate it in the tracking issue and execute it (i.e. by reverting the PR or making sure the author fixes forward).
+The dev workflow leads are developers with revert permissions, so the above instructions apply to you too. Further, leads should make sure a tracking issue is created and alert the rest of the dev workflow leads. The dev workflow leads are also responsible for making sure whatever stragety we decide on to fix develop is actually executed.
```
seanlip: Bit confusing to have the switch between second/third person here (e.g. "you" in first sentence), assuming "you" and "dev workflow leads" mean the same thing.

seanlip (commented):
Thanks @U8NWXD, left some comments.

One question -- would it be OK to add some examples about what "severe" (say) means? E.g. if it blocks them making a PR because CI checks are failing on develop, it's probably severe. Maybe some examples of "less severe" things may be helpful too (doesn't need to be exhaustive but can give folks an idea for common cases).

Revert-and-Regression-Policy.md
U8NWXD: The dev workflow leads are responsible for considering mitigations to prevent a similar regression from happening again. This is discussed in the dev workflow leads section. I'll remove that bit though since it's not necessary

Revert-and-Regression-Policy.md
U8NWXD: fixed

Revert-and-Regression-Policy.md
U8NWXD: fixed

Revert-and-Regression-Policy.md
```diff
@@ -1,14 +1,22 @@
...
   * [Dev Workflow Leads](#dev-workflow-leads)
 * [Tips](#tips)
   * [How to identify the bad commit](#how-to-identify-the-bad-commit)
   * [How to revert a PR](#how-to-revert-a-pr)
 
+## Blame-Free Reverts
+
+Sometimes, PRs need to be reverted despite the PR author doing nothing wrong. For example, sometimes a race condition occurs where 2 PRs are merged in quick succession and conflict. We may have to revert one of the PRs to fix develop even though no PR author did anything wrong.
+
+It is important to note that reverts are not ment to be punitive. A key tenet of the dev workflow team is that develop needs to be kept correct at all times. When a regression appears, we need to fix develop as quickly as possible to avoid blocking other developers, and sometimes the quickest way is to revert PRs. Thus, we revert PRs to help developers keep working, not because the PR author did something wrong.
```
seanlip: ment --> meant

seanlip (approved):
LGTM, just one typo!

Revert-and-Regression-Policy.md
U8NWXD: fixed

Issue #97, opened 2023-01-23
Include Mypy in CI checks


SD-13:
Add Mypy in the CI to check for type annotations in the wiki repo.

Suggested [here](https://github.com/oppia/oppia-web-developer-docs/pull/84/#discussion_r1083556925) by @U8NWXD

U8NWXD:
We should enforce mypy strictly (i.e. require that everything be typed and that there are no type errors). Checks should be easy to run locally, be run on each commit to develop, and be enforced for all PRs to develop

deadex-ng:
@gp201 I would like to work on this

gp201:
@deadex-ng I have assigned this issue to you.

deadex-ng:
@gp201 Should the mypy workflow check all python files or it should only check python files that have been changed when a PR is made?

U8NWXD:
We need to check all Python files because a change in one file could cause a type error in another (unmodified) file

Shivkant-Chauhan:
@deadex-ng please update the project board with the expected merge date.

deadex-ng:
@Shivkant-Chauhan I made a [PR](https://github.com/oppia/oppia-web-developer-docs/pull/119) and it was merged.

Shivkant-Chauhan:
Closing this as completed by #119

PR #119, opened 2023-02-23
Include Mypy in CI checks

deadex-ng:
@gp201 - The PR adds mypy check to the workflow, according to issue [#97](https://github.com/oppia/oppia-web-developer-docs/issues/97)
The workflow checks all the python files in the repository, both modified and unmodified files.

### DIFF
--- a/.github/workflows/mypy_check.yml
+++ b/.github/workflows/mypy_check.yml
@@ -1,0 +1,15 @@
+name: "mypy check"
+on: [pull_request]
+
+jobs:
+
+  static-type-check:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-python@v3
+      with:
+        python-version: '3.x'
+    - run: pip install mypy
+    - name: Run mypy check on all files
+      run: mypy --strict .
--- a/.github/workflows/mypy_check.yml
+++ b/.github/workflows/mypy_check.yml
@@ -1,6 +1,8 @@
 name: "mypy check"
-on: [pull_request]
-
+on:
+  pull_request:
+    branches:
+      - develop
 jobs:
 
   static-type-check:
--- a/.github/workflows/mypy_check.yml
+++ b/.github/workflows/mypy_check.yml
@@ -1,6 +1,6 @@
 name: "mypy check"
 on:
-  pull_request:
+  push:
     branches:
       - develop
 jobs:
--- a/.github/workflows/mypy_check.yml
+++ b/.github/workflows/mypy_check.yml
@@ -3,6 +3,9 @@ on:
   push:
     branches:
       - develop
+  pull_request:
+    branches:
+      - develop
 jobs:
 
   static-type-check:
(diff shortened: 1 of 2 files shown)


U8NWXD:
Thanks @deadex-ng! Looks like there are some type errors. Could you fix those?

deadex-ng:
> Thanks @deadex-ng! Looks like there are some type errors. Could you fix those?

@U8NWXD  Sure, I can fix those.

.github/workflows/mypy_check.yml
```diff
@@ -0,0 +1,15 @@
...
+
+jobs:
+
+  static-type-check:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-python@v3
+      with:
+        python-version: '3.x'
```
U8NWXD: Can we pick a specific version of Python (say 3.8 to match oppia-web)? We like pinning all our versions so that an upgrade doesn't surprise us and break something

.github/workflows/mypy_check.yml
```diff
@@ -0,0 +1,15 @@
+name: "mypy check"
+on: [pull_request]
```
U8NWXD: Can we run this on develop too? That helps just in case 2 PRs get merged in quick succession and their changes have clashing types

U8NWXD (changes requested):

.github/workflows/mypy_check.yml
deadex-ng: Yes, we can run it on develop.

.github/workflows/mypy_check.yml
deadex-ng: I have specified the version, 3.8.

.github/workflows/mypy_check.yml
U8NWXD: I think you need to change the workflow so that it runs on develop. For example, here's how we do it in oppia/oppia:

https://github.com/oppia/oppia/blob/24298204cfaf7ade44fcf7542418cc1d66de77a1/.github/workflows/backend_tests.yml#L5-L12

deadex-ng: I have made the changes, it now runs on develop and thank you for the resource.

U8NWXD: Hmm I think now it runs on pull requests opened against `develop`, but I think you still need the `push` key to tell it to run whenever a commit gets pushed to develop (i.e. whenever a PR merges). You can read GitHub's documentation here: https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#using-events-to-trigger-workflows

deadex-ng: Thank you for the resource. I have added the `push` event.

U8NWXD (approved):

SD-13:
Hey @U8NWXD, just realized that we are using the checkout v3 in all workflows in the oppia repo. So, what do you think about following the same practice here? It may be a nit but just wanted to know, do we care about this?

PR #197, opened 2023-10-06
Fix #196: Adding instructions window users + updating troubleshooting page

### LINKED ISSUE #196: Docker Setup wiki instructions not clear for Windows users, add issue in Troubleshooting page
Docker Setup Wiki must include instructions for installing `make` for Windows users. 
Some users are facing `Operation not permitted` issue in the docker setup, and using `sudo` with the make command fixes that. Include this information in the Troubleshooting page.

Shivkant-Chauhan:
Wiki modifications for #196 
Added instructions for install `make` in Windows + updating troubleshooting page for the admin access issue


Installing-Oppia-using-Docker.md
```diff
@@ -74,6 +74,7 @@ To install Oppia under Docker, follow these steps:
 ### Pre-requisites for setting up the development server
 
 1. **Install Docker Desktop**: Download and install the latest version of Docker Desktop from the [official Docker website](https://www.docker.com/products/docker-desktop/). Docker Desktop provides a user-friendly interface for developers. Follow the steps in the given link to download and install it.
+2. **[For Windows users]**: The Windows users must have `make` installed on their system. To install `make`, follow the steps given in the following link: [Installing make on Windows](https://stackoverflow.com/questions/32127524/how-to-install-and-use-make-in-windows).
```
seanlip: The Windows users --> You
their --> your

Can you explain how a developer can verify they have `make` on their system? The flow here should be "do this to see if you have it; if you don't, follow these instructions".

Troubleshooting.md
```diff
@@ -688,6 +690,21 @@ Following the completion of the aforementioned commands, you can proceed with th
...
+```
+Operation not permitted
+make: *** [Makefile:14: docker] Error 1
+```
+
+The simple solution for the above error is to run the make commands with admin access i.e. with `sudo`, example:
+```
+sudo make build
+```
+The cause of the issue is that Docker daemon binds to a unix socker by default, which is owned by root. So, when you run the make commands without admin access, it is not able to access the unix socket and hence the error is thrown.
```
seanlip: The first "socker" is misspelled

Also, in what circumstances does this happen? Does it happen only for particular OS's? What differentiates machines on which `sudo` is necessary and on which it is not? (See comment above too, re trying to avoid this altogether.)

Troubleshooting.md
```diff
@@ -688,6 +690,21 @@ Following the completion of the aforementioned commands, you can proceed with th
...
 
+### make commands: `Operation not permitted`
+
+Some users might face the following error while running `make` commands:
+```
+Operation not permitted
+make: *** [Makefile:14: docker] Error 1
+```
+
+The simple solution for the above error is to run the make commands with admin access i.e. with `sudo`, example:
```
seanlip: Users will likely want to avoid running sudo where possible, because this gives the script elevated permissions. Can we instead fix things so that they don't need to do this? E.g. https://askubuntu.com/questions/477551/how-can-i-use-docker-without-sudo

seanlip (changes requested):
Thanks @Shivkant-Chauhan, left some comments.

Troubleshooting.md
Shivkant-Chauhan: done. PTAL

Troubleshooting.md
Shivkant-Chauhan: it happens when the current user donot have root priviledge for accessing the UNIX socket. instead of adding `sudo` since it add evevated permissions, we can simply add a new docker user group. this fixes the issue for linux and macOS users. not is if this happens for windows also as no one has encountered such issue

Installing-Oppia-using-Docker.md
Shivkant-Chauhan: done. PTAL

Shivkant-Chauhan:
@seanlip PTAL. thanks!

Troubleshooting.md
```diff
@@ -688,6 +690,25 @@ Following the completion of the aforementioned commands, you can proceed with th
...
 
+### make commands: `Operation not permitted`
+
+Some users might face the following error while running `make` commands:
+```
+Operation not permitted
+make: *** [Makefile:14: docker] Error 1
+```
+
+Create a new user called 'docker' and include the current user in the 'docker' group. Execute the following commands to achieve this:
```
seanlip: Create -->  To fix this, create

seanlip (approved):
Just one note, otherwise lgtm. Thanks!

Troubleshooting.md
Shivkant-Chauhan: done

seanlip:
Thanks, LGTM!

PR #199, opened 2023-10-14
fix link to the Editor-pages guide

nikitaevg:
Currently this link forwards to the raw content (to [here](https://raw.githubusercontent.com/wiki/oppia/oppia/Editor-pages.md#steps-to-create-a-topicskill)). I believe it'd be better if it led to an HTML. As per [this](https://stackoverflow.com/questions/34804531/in-github-wiki-absolute-links-to-markdown-leads-to-raw-githubusercontent) question we don't need .md in the link

### DIFF
--- a/LaCE-onboarding-guide.md
+++ b/LaCE-onboarding-guide.md
@@ -10,7 +10,7 @@ This page will help you solve common queries that you may face while working on 
 | 4.    | [Creating a classroom](#4-creating-a-classroom)|
 
 ## 1. Creating a Topic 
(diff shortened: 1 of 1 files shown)


nikitaevg:
Here's the video of what I mean. I'm not sure how to verify that it works though, because link resolution works differently in wiki and if viewing files in the repository.
[Screencast from 14.10.2023 12:00:58.webm](https://github.com/oppia/oppia-web-developer-docs/assets/6707908/d77b1d24-aa03-44fe-af2d-19cd80c207c1)

seanlip (approved):
Thanks, I think this should be fine. Not sure how to test it but feel free to make another PR if it still looks bad after this one is merged!

PR #217, opened 2023-12-27
Fixed broken github blame image

jaiakash:
Hi i have added the image for GitHub blame image that was broken in the wiki.

Closes #215

PR #223, opened 2024-01-03
Added domain block fix for Jio users

jaiakash:
Some of the GitHub domains are blocked on the Jio network in India, so here I added the fix. 
See related discussion #217


seanlip (approved):
Looks great. Thanks @jaiakash! LGTM.

PR #268, opened 2024-03-03
Make a separate page for Testing Emails.

Ash-2k3:


Testing-Email-functionality.md
```diff
@@ -0,0 +1,19 @@
+To test the email functionality of Oppia, you can follow these steps:
+
+## Testing with App Engine Deployment
```
seanlip, line 3: @Ash-2k3 This is a very heavyweight way to test. I suggest swapping the two sections so that the local version comes first, and not calling the local version a "workaround". In fact I think I might suggest dropping the first section altogether since that is quite an impractical way of testing.

Also could you please verify the local version of the instructions and update as needed? The given file doesn't exist in the codebase any more, and I think the dev_mode_email_services.py should "just work" out of the box. I think it would be helpful to walk through this with a specific example with screenshots, since many devs don't seem to know how to use the local server terminal to read logs...

seanlip (changes requested):
Thanks @Ash-2k3! Left a comment, PTAL.

Ash-2k3:
I had to re-fork the repo as my main branch got dirty and it would have taken some work to get it clean again, thus I took the shortcut. I will address @seanlip's comment in a new PR which I plan to open by March end. (As I'll be going off soon in next two days)

PR #285, opened 2024-03-20
Rename heading for filing a bug report pointer.

Ash-2k3:


_Sidebar.md
```diff
@@ -9,7 +9,7 @@
     * [[Voice artists | Instructions-for-voice-artists]]
     * [[Designers and artists|Contributing-to-Oppia's-design]]
   * Reporting a bug
-    * [[File an issue for bug|Github-Issue-Guide]]
+    * [[How to File a Bug Report|Github-Issue-Guide]]
```
seanlip: Why does this need to be in a sub-bullet point vs at the top-level?

Also should the page name be renamed to match the link title? "GitHub Issue Guide" is a bit vague and if the link and destination don't match, users may think they navigated to the wrong place.

seanlip (changes requested):

_Sidebar.md
```diff
@@ -9,7 +9,7 @@
     * [[Voice artists | Instructions-for-voice-artists]]
     * [[Designers and artists|Contributing-to-Oppia's-design]]
   * Reporting a bug
-    * [[File an issue for bug|Github-Issue-Guide]]
```
Ash-2k3, line 12: I remember now why I named it like this; For Lesson creation guide, it's named as `Create a Lesson!`, so I thought this would make more sense.

seanlip: I think that was intentionally phrased as an invitation, but here we aren't exactly inviting folks to file reports, just providing the guide for how to do it if they need to.

_Sidebar.md
Ash-2k3: I have made it as top-level point now and also renamed it to match the link title. Thanks !

_Sidebar.md
Ash-2k3: Got it, Thanks!

seanlip (approved):
Thanks Ashwath!

PR #293, opened 2024-03-27
Fix SoD 2024 TOC and Headings, and Add Pre-Submission Review Instructions

U8NWXD:
This PR makes the following changes to the SoD 2024 page:
* Fix TOC links
* Fix headings to avoid using h1 (since GitHub automatically uses the file name to generate an h1 title)
* Add instructions for pre-submission review
* Add the SoD admins email for confidential contact


seanlip (approved):
Thanks @U8NWXD !

PR #306, opened 2024-04-07
Fix #302: Mentioned typical time for tests (frontend and acceptance tests) executions in respective docs

### LINKED ISSUE #302: Document Typical Test Execution Duration of all type of tests.
I observed that the frontend test execution on my MAC M1, was taking around 30-40 minutes. I initially thought this to be normal due to a lack of documentation specifying the expected duration. However, I later discovered that this was due to the lack of optimization of Chrome for arm64 (it’s recently been fixed, possibly because of the new release of Chrome). On other devices (may not be all), the tests typically run for about 3-5 minutes.

For the benefit of newcomers and to prevent any confusion in the future, it may be a good idea to document the typical test execution time of all type of tests we can run locally, assuming sufficient resources are available. This will provide a benchmark for what is considered ‘normal’, allowing developers to identify any potential issues or anomalies in test execution time.

Please note that the actual duration may vary based on the specific test and device resources.

Akhilesh-max:
Fix #302

### DIFF
--- a/Acceptance-Tests.md
+++ b/Acceptance-Tests.md
@@ -77,7 +77,7 @@ Docker:
 ```
 make run_tests.acceptance suite="blog-editor-tests/check-blog-editor-unable-to-publish-duplicate-blog-post"
 ```
-
+**Note: Typically, these tests take anywhere between 0.5 to 2-3 minutes (excluding the time taken for setting up the server) for any suite to run, both in headless and non-headless modes, assuming the machine has sufficient resources. If the total time taken is significantly longer than this, it may indicate an issue with the testing environment. If such an issue is noticed or observed, please raise it on our [issue tracker](https://github.com/oppia/oppia/issues).**
 
 ## How to write new tests for a specific user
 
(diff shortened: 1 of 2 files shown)


Akhilesh-max:
@seanlip, I have mentioned the typical time it may take for frontend and acceptance tests to run across various devices.
However, I haven’t done the same for e2e tests as these tests vary significantly based on the available computing resources. For instance, it took Hardik 6 minutes to run a suit `topicAndStoryEditor` on a Windows machine with an i7 processor and 16 GB of RAM, whereas it took me 10 minutes to run the same suit on a MAC M1 with 8 GB of RAM. Ashwath, who has the same machine configuration as mine, took 16 minutes as he was simultaneously working on multiple other tasks. Given this variability, I’m uncertain whether we should provide a typical execution time range for these tests. Please suggest. 

Also, lighthouse tests are not passing locally across devices as per this [issue](https://github.com/oppia/oppia/issues/19909). So, couldn't really mention for that as well. 

Thanks.

seanlip (approved):
This looks pretty good. Thanks!

seanlip:
@Akhilesh-max I understand the limitations, thanks for going into detail about those. I think it's fine to omit them for now.

Thanks!

PR #344, opened 2024-07-09
Added patch file for first tutorial: Tutorial debug server errors

Ash-2k3:


Ash-2k3:
Closing this as we are not using patch file instead letting the folks checkout to a commit

PR #345, opened 2024-07-22
Update Installing-Oppia-(Mac-OS;-Python-3).md

sezavala:
The previous instructions would always lead to an error on my mac M3, stating "use_python: command not found", even though I had the script in my .direnv file. After reviewing the direnv documentation, I discovered an article where they were using a ruby script, but they added it to the ~/.config/direnv/direnvrc file. After adding my script to the file, direnv was able to load the script and I was able to run the build! Hopefully this can help any other mac users with the same issue! @seanlip @gp201 
Screenshot 2024-07-22 at 12 47 05 PM

### DIFF
--- a/Installing-Oppia-(Mac-OS;-Python-3).md
+++ b/Installing-Oppia-(Mac-OS;-Python-3).md
@@ -68,7 +68,7 @@ Oppia relies on a number of programs and third-party libraries. Many of these li
 
 2. Install [direnv](https://direnv.net/).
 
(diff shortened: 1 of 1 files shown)


sezavala:
Also for reference, here is the documentation I found: https://direnv.net/docs/ruby.html

seanlip (approved):
Thanks @sezavala! Sorry for the delay in looking at this, I was travelling recently and am catching up on backlog. LGTM.

PR #389, opened 2024-09-18
Fix: add space to fix Docker instructions

annhe387:
Fixes the SSH config example by adding a space between Host and *. This prevents the following error after copying it into the .ssh/config file:
/Users/xxx/.ssh/config line 1: no argument after keyword "host*"
/Users/xxx/.ssh/config: terminating, 1 bad configuration options


seanlip (approved):
Thanks very much for the fix, @annhe387. LGTM!

PR #400, opened 2024-11-01
Changed all instances of Python 3.8.x and replaced with 3.9

ProfessionalCaddie:
This is step 2 in https://github.com/oppia/oppia/issues/21020


Installing-Oppia-(Linux;-Python-3).md
```diff
@@ -140,15 +140,15 @@ For your virtual environment, we recommend you use [pyenv](https://github.com/py
 ```bash
 exec "$SHELL"
 ```
-4. Now you can install Python 3.8.15 and the associated pip like this:
+4. Now you can install Python 3.9 and the associated pip like this:
```
seanlip: Hi @ProfessionalCaddie, thanks for your PR!

Just a note, for anywhere there was a patch version originally (3.8.15), please replace with 3.9,20 instead of just 3.9. This is for consistency, since we need to make sure that everyone is using the same specific version of Python that is used in the production server (cf https://cloud.google.com/appengine/docs/standard/python3/runtime#python_3_versions). Thanks!

seanlip (changes requested):

Installing-Oppia-(Linux;-Python-3).md
```diff
@@ -140,15 +140,15 @@ For your virtual environment, we recommend you use [pyenv](https://github.com/py
...
 
    ```console
-   $ pyenv install 3.8.15
-   installing python-3.8.15...
+   $ pyenv install 3.9
+   installing python-3.9...
    patching file misc/news.d/next/build/2021-10-11-16-27-38.bpo-45405.isfdw5.rst
```
seanlip: I think you can drop this line and the next two as well.

ProfessionalCaddie: @seanlip, I took your comment literally. I erased lines 148-150. If this is not what you meant please let me know.

seanlip: Yes that is correct. When I followed these steps I didn't get the output on lines 148-150.

seanlip (approved):
Thanks @ProfessionalCaddie. Looks good!

PR #414, opened 2024-11-16
Consolidate i18n pages and simplify them to explain the key things a developer would need to know.

seanlip:
(Also, add footprints to tutorials in the sidebar.)


seanlip:
@Ash-2k3 @U8NWXD PTAL. Thanks.

Adding-new-translations-for-i18n.md
```diff
@@ -1,14 +1,33 @@
...
 3. Select a language to contribute translations for. (Also, read the [notes](https://github.com/oppia/oppia/wiki/Adding-new-translations-for-i18n#note-1-variable-replacement) below describing the translation formats used for variables and plurals.)
 
-Changes will then be pushed to Oppia automatically by the translatewiki admins (**@Nikerabbit** and **@siebrand**), and they will show up in future Oppia releases.
+Changes will then be pushed to Oppia automatically by the Translatewiki admins, and they will show up in future Oppia releases. We typically update new translations to the Oppia.org website on a monthly cadence.
+
+### Important: Don't rely on machine translation
+
+The translatewiki admins have requested that translators do not rely on machine auto-translation, especially if they don't know the language and cannot fix its mistakes. 
+
+In cases where filling in a translation is necessary for technical reasons, they recommend doing the following:
```
U8NWXD: I'm unclear on when machine-generated translations are acceptable. Is the use-case here that I need to provide a translation for technical reasons, but I can't do the translation personally?

Adding-new-translations-for-i18n.md
```diff
@@ -1,14 +1,33 @@
...
+- Add the string `!!FUZZY!!` to the beginning of the machine-translated string. This will update the translation in Oppia's source tree, and also mark it as "needing update" for the translators on translatewiki.
```
U8NWXD, line 29: Do we have technical safeguards to block these fuzzy-labeled translations from being shown to users in the time before they get manually translated?

Adding-new-translations-for-i18n.md
```diff
@@ -33,12 +52,138 @@ In the example above, if there is only one notification, then we should change "
...
+
+### Good i18n practices
+
+Please consider i18n while developing. Not all languages are the same: words have different lengths, pluralization rules differ, sentences have different structures and the direction of writing can be from right to left. As a result, sometimes development practices that are generally good (like code reuse) turn out to be less than ideal for i18n. Here are some important points to take into account:
+- Do not include raw strings, always use a translation key. Even if it is just an exclamation mark (!)
+- When designing a page, plan for the case where strings are twice the length in other languages. Also, think about how the page would look like if the language is written from right to left.
+- Try to include as much text as you can in a single key, so that the translator can provide a more coherent translation. Do not divide a paragraph into multiple keys unless you cannot avoid it. Don't split strings up and concatenate them, since different languages will use a different grammatical order.
+- If you need to include html in your string, such as `<a></a>` tags, try and pass the code as an argument to the translation service.
```
U8NWXD, line 72: how do we do this? an example might help

Adding-new-translations-for-i18n.md
```diff
@@ -33,12 +52,138 @@ In the example above, if there is only one notification, then we should change "
...
+- Add the translation key inside the html tag as the value of the translate attribute. This will prevent the key from being shown briefly -- instead, the location of the string will remain empty in the interim.
+- If the string is in a very visible location and there is FoUC in the preferred language, add the key and the translation into the `DEFAULT_TRANSLATIONS` constant defined in the file [i18n.js](https://github.com/oppia/oppia/blob/develop/core/templates/i18n.js).
+
+### Titles
+
+To add a title to the window, replace the block maintitle for a translation key:
+
+    {% block maintitle %}TRANSLATE_KEY{% endblock maintitle %}
+
+and add the translation into the corresponding json files. Titles should always be translated, but this is not mandatory. If you replace this block with a common string, it will be shown as the title instead.
```
U8NWXD: why do we allow untranslated titles?

U8NWXD (changes requested):

Adding-new-translations-for-i18n.md
seanlip: Yes, that's correct. The case in which this arose was when a translatewiki string was wrong and it was breaking tests. It's fairly rare though.

I updated the doc to clarify, thanks for pointing this out.

Adding-new-translations-for-i18n.md
seanlip: We don't, but that's OK IMO. A not-so-great translation is better than no translation.

Adding-new-translations-for-i18n.md
seanlip: This was mentioned above, so I added a reference link.

Adding-new-translations-for-i18n.md
seanlip: I actually just copied this from the previous doc, but ... wow, this is really old, this is the Jinja syntax from, like, 2016 :P I deleted it entirely since it no longer applies!

seanlip (commented):
Thanks @U8NWXD, I addressed the comments. PTAL!

U8NWXD (approved):

PR #419, opened 2024-11-19
Fix #387: Clean up backend type annotations page.

### LINKED ISSUE #387: Mypy Denylist 'NOT_FULLY_COVERED_FILES' Referenced but Not Present
There seems to be an error in the Oppia Wiki regarding the Mypy checks configuration. According to the [Backend Type Annotations Wiki Page](https://github.com/oppia/oppia/wiki/Backend-Type-Annotations#when-to-add-type-annotations), there should be a denylist called `NOT_FULLY_COVERED_FILES` inside `scripts/run_mypy_checks.py`. However, this denylist is not present in the codebase, which implies that all files are currently covered by Mypy checks.

Thus we would want to update the wiki to make sure wiki correctly reflects the current configuration of Mypy checks.

seanlip:
This fixes https://github.com/oppia/oppia-web-developer-docs//issues/387 by removing info on the wiki page related to the no-longer-existent mypy denylist.

### DIFF
--- a/Backend-Type-Annotations.md
+++ b/Backend-Type-Annotations.md
@@ -2,15 +2,7 @@
 
 Type Annotations are a new feature added in [PEP 484](https://www.python.org/dev/peps/pep-0484/) that allow for adding type hints to variables. They give information about types of variables to someone who is reading the code. This brings a sense of statically-typed control to the dynamically typed Python. Though Python ignores these type hints during code execution, third-party libraries can be used to statically type-check the codebase.
 
-
-## When to add type annotations
-We use mypy to type check the backend part of our codebase. A file becomes eligible for type checking if it falls under one of the following 3 cases:
-1. If a **new file** is being added, it must be fully type annotated and hence type checked.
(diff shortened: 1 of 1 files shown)


Ash-2k3 (commented):
Thanks @seanlip. Just one thought - Should we also mention the limitation of the mypy and when it might not work as expected ?

Something like - "mypy is a static type checker for Python, designed to enforce type safety by verifying that types declared in type annotations are respected throughout the code. However, mypy has limitations when dynamic typing is involved, especially with the use of functions like `setattr` that dynamically set attributes at runtime."

seanlip:
@Ash-2k3 Hm, I think not. Outside tests, we only use setattr in two places and I don't think we should normalize it. If anything I think we should get rid of the usage of setattr.

Ash-2k3:
Okay, in that case, let's file a bug to drop the usage of setattr, as discussed in our Google Chat conversation.

seanlip:
@Ash-2k3 Done -- https://github.com/oppia/oppia/issues/21293.

Ash-2k3 (approved):
Thanks, LGTM!

PR #426, opened 2024-12-22
Tutorial on how to write migration jobs

### LINKED ISSUE #364: Create Tutorial - Learn how to safely make a backend change that has an impact on server data.
## Skills covered

Beam Jobs(Audit Job + Migration Job), Define and document the launch process,Test-run the Beam job and fix issues.

## Example

Implement a restriction on the “Bio” field in the user’s Preferences page that limits what is entered in the field to a maximum of 200 characters.

Ash-2k3:
Fix https://github.com/oppia/oppia-web-developer-docs/issues/364

Link to Google doc - https://docs.google.com/document/d/1kOTBlrrCKu436A7pL4KnzWYHGoBYTxYhVI-VG3pRMJo/edit?tab=t.0


Ash-2k3:
@seanlip PTAL at this PR, thanks!

Tutorial-Learn-how-to-make-a-backend-change-that-has-an-impact-on-server-data.md
```diff
@@ -0,0 +1,677 @@
...
+
+# Introduction
+
+In this tutorial, you will learn how to safely implement backend changes that impact server data, an essential skill for any developer working with applications that store data. We’ll cover key concepts like modifying data models, writing and testing Beam Jobs (Audit and Migration), and documenting a reliable launch process. These skills are fundamental for maintaining data integrity, ensuring data consistency, and avoiding disruptions during backend updates.
+
+By the end of this tutorial, you will have the knowledge and confidence to handle backend changes at Oppia effectively, ensuring data safety and smooth application performance.
+
+# Scenario
+
+In this tutorial, we will address an issue where the `user_bio` field in `UserSettingsModel` allows users to enter bios of unrestricted length. While this provided users with flexibility in expressing themselves, it has become necessary to enforce a length limit of 200 characters. This change ensures consistency and allows UI designers to reliably allocate space for displaying bios, improving the overall user experience.
```
seanlip: Change the "While this ... overall user experience." part to "For the purposes of this tutorial, imagine that the technical team has decided to enforce a length limit of 200 characters for this field, in order to ensure consistency and allow UI designers to reliably allocate space for displaying bios."

Tutorial-Learn-how-to-make-a-backend-change-that-has-an-impact-on-server-data.md
```diff
@@ -0,0 +1,677 @@
...
+# Procedure
+
+## Section 1: Navigating and Understanding the Preferences Page
+
+Start by navigating to the Preferences page in your local development environment.  
+**URL**: [`http://localhost:8181/preferences`](http://localhost:8181/preferences)
+
+Our goal is to identify which storage model stores the fields shown on this page. There are multiple ways to approach this:
+
+1. **Code Exploration**: Review the relevant files (e.g., `gae_models.py`) to manually inspect the fields. Keep in mind that this approach may not be ideal for new contributors who are just beginning to familiarize themselves with the codebase. It is typically more suitable for those who have already spent significant time working with and understanding the structure of the codebase.
```
seanlip: You are switching suddenly from first/second-person to third-person here. It might be better to address the contributor directly, e.g. "Note that, if you're new to the codebase, ...".

Also I'm not really sure you even need the first point here.  I think you can say something like "...which storage model stores the fields shown on this page. If you know which one it is already, you can inspect the fields directly. Otherwise, here is how to find out ..." (feel free to reword/simplify as needed).

Tutorial-Learn-how-to-make-a-backend-change-that-has-an-impact-on-server-data.md
```diff
@@ -0,0 +1,677 @@
...
+
+Before you begin, ensure that you have completed the following steps to set up and prepare your development environment:
+
+1. Set Up Oppia Locally: Follow the [Oppia setup instructions](https://github.com/oppia/oppia/wiki) to clone the repository and set up your local development environment.  
+2. Familiarize Yourself with Apache Beam Jobs: [Beam jobs](https://github.com/oppia/oppia/wiki/Apache-Beam-Jobs) are an integral part of data processing at Oppia. Before proceeding, take some time to understand their role in auditing and migrating data. You can refer to the [Apache Beam Jobs at Oppia tutorial](https://github.com/oppia/oppia/wiki/Tutorial-Learn-How-to-Write-and-Test-a-Non-Trivial-Beam-Job) for detailed guidance on writing and testing Beam jobs.  
+3. Understand the Preferences Page: This tutorial involves modifying the Bio field in the user’s Preferences page. To understand the context better, go through the [How to Access Oppia Webpages: Preferences Page](https://github.com/oppia/oppia/wiki/How-to-access-Oppia-webpages#preferences-page) guide.
+
+# Procedure
+
+## Section 1: Navigating and Understanding the Preferences Page
```
seanlip: Give this a clearer title that makes it clear when the section is "done". How would the developer know that the 
"understanding" part has been completed?

Tutorial-Learn-how-to-make-a-backend-change-that-has-an-impact-on-server-data.md
```diff
@@ -0,0 +1,677 @@
...
+
+1. **Code Exploration**: Review the relevant files (e.g., `gae_models.py`) to manually inspect the fields. Keep in mind that this approach may not be ideal for new contributors who are just beginning to familiarize themselves with the codebase. It is typically more suitable for those who have already spent significant time working with and understanding the structure of the codebase.  
+2. **Network Inspection**: Observe the network call triggered when changes are made on the Preferences page to trace the update process through the codebase.
+
+In this tutorial, we’ll use the second approach to showcase how to trace functionality effectively.
+
+> [!IMPORTANT]
+> Practice 1: Familiarize yourself with the codebase architecture at Oppia. Understanding the structure of the codebase will help you navigate through various layers of code at Oppia more efficiently. Follow this guide: [Overview of the Oppia Codebase](https://github.com/oppia/oppia/wiki/Overview-of-the-Oppia-codebase).
+
+![Screenshot of Preference Page](images/TutorialMigrationJob/preferencePage.png)
```
seanlip: Both in the alt text and image filename, I think it's the "Preferences" page.

Tutorial-Learn-how-to-make-a-backend-change-that-has-an-impact-on-server-data.md
```diff
@@ -0,0 +1,677 @@
...
+
+This tells us that the endpoint (`/preferenceshandler/data`) handles updates to the Preferences page.
+
+![Screenshot of Preference Page with Network Tab opened](images/TutorialMigrationJob/preferencePageWithNetworkTab.png)
+
+Now it’s time to trace this endpoint in the codebase.
+
+The URL triggered when clicking the **Save Changes** button is `/preferenceshandler/data`. Perform a quick search in the codebase for this endpoint. Note that exact URL matches may not yield results, as URLs are often aliased as constants. To locate the relevant code, try variations such as `/preferenceshandler/data`, `/preferenceshandler`, or `preferenceshandler`.
+
+> [!IMPORTANT]
```
seanlip, line 99: I think you should explain something before this practice about main.py containing all the routings, so it's less about "generally search the codebase" and more about "easily find the relevant routing in main.py". It's great that you have the note about the possible aliasing-by-constants.

Tutorial-Learn-how-to-make-a-backend-change-that-has-an-impact-on-server-data.md
```diff
@@ -0,0 +1,677 @@
...
+In `feconf.py`, add:
+
+```python
+MAX_BIO_LENGTH_IN_CHARS = 200
+```
+
+Update the `put` Method in the Controller by adding a check to enforce the character limit before saving: 
+
+```python
+elif update_type == 'user_bio':
```
seanlip, line 222: Just to check, is this supposed to be indented so far to the left?

Tutorial-Learn-how-to-make-a-backend-change-that-has-an-impact-on-server-data.md
```diff
@@ -0,0 +1,677 @@
...
+               self.__validate_data_type(update_type, str, data)
+               if len(data) > feconf.MAX_BIO_LENGTH_IN_CHARS:
+                   raise self.InvalidInputException(
+                       'User bio exceeds maximum character limit: %s'
+                       % feconf.MAX_BIO_LENGTH_IN_CHARS)
+               user_settings.user_bio = data
+```
+
+***Note**: Normally, we would use schema validation to enforce this (e.g., using a regex as demonstrated [here](https://github.com/oppia/oppia/blob/cac148abaaa0bba4d96b9df26aa67fd3068b216c/core/schema_utils.py#L700)). However, the `preferences` handler hasn’t been set up for schema validation yet. Adding schema validation would require defining validations for the entire handler, which is beyond the scope of this tutorial.*
```
seanlip: Would it be better to link to https://github.com/oppia/oppia/wiki/Schemas#how-to-write-validation-schemas-for-handlers instead (updating the surrounding documentation as needed)?

Tutorial-Learn-how-to-make-a-backend-change-that-has-an-impact-on-server-data.md
```diff
@@ -0,0 +1,677 @@
...
+
+For example, potential options might include:
+
+1. Truncating the `user_bio` field to the maximum allowed length (200 characters).  
+2. Clearing the `user_bio` field entirely for users with long bios.  
+3. Leaving the bios unchanged but marking them as needing a manual update.
+
+For this tutorial, we will choose the **truncation** approach, ensuring all user bios conform to the 200-character limit.
+
+***Note**: While this approach is simple and sufficient for the purposes of this tutorial, it may lead to a sub-optimal user experience as it truncates user input without providing feedback or allowing edits. This is not necessarily a best practice for real-world applications but serves as an illustrative example for learning.*
```
seanlip: I think you'd want to explain that, in practice, for decisions like this that aren't clear, the developer should compile a list of the different options and their pros/cons and then discuss with the product team and technical leads what the best thing to do is (for changes that affect user experience). Then you can explain separately that, for the purposes of this tutorial, the team leads chose option 1.

Also, why is there a trailing * at the end of this paragraph? There are also three stars before "Note" and two after it.

Would it also be better to use things like https://github.com/orgs/community/discussions/16925 for notes?

Tutorial-Learn-how-to-make-a-backend-change-that-has-an-impact-on-server-data.md
```diff
@@ -0,0 +1,677 @@
...
+
+Visualizing our Beam job as a DAG helps in planning the structure and flow of our data processing pipeline. It provides a clear picture of how data moves from one step to another, ensuring that all necessary operations are performed in the correct order.
+
+The Beam job’s objective is to truncate the `user_bio` field in the `UserSettingsModel` datastore records exceeding the 200-character limit. The workflow can be broken down into the following steps:
+
+> [!IMPORTANT]
+> Practice 8: Take a notebook and try drafting a rough workflow of what our job would do, using boxes for the steps and arrows to connect different steps. 
+> 
+> Hint: 
+> - Read Everything First: Start by reading all the necessary data at the beginning of the job. This ensures that you have all the required information before performing any operations.
```
seanlip: Make "Read Everything First." bold and put a period at the end of it, instead of a colon. Similarly for the other bullet points.

Tutorial-Learn-how-to-make-a-backend-change-that-has-an-impact-on-server-data.md
```diff
@@ -0,0 +1,677 @@
...
+
+#### **Thought Process for the Audit Job**
+
+1. **Simulating Logic**: The audit job must simulate the exact same steps as the main Beam job to ensure consistency in logic and results.  
+2. **Read-Only Operations**: Unlike the main job, an audit job should not persist any changes to the datastore. This avoids unintended side effects during testing.  
+3. **Detailed Reporting**: The job should generate a detailed report or log indicating the records that require updates. This transparency helps validate the scope and correctness of the job.  
+4. **Reusable Patterns**: Follow established patterns and conventions for audit jobs in the Oppia codebase.
+
+> [!IMPORTANT]
+> Practice 9: Based on the above explanation and thought process. Can you write down the Audit Job for our use case.
```
seanlip: The first sentence of this is a sentence fragment; please update.

Tutorial-Learn-how-to-make-a-backend-change-that-has-an-impact-on-server-data.md
```diff
@@ -0,0 +1,677 @@
...
+
+For instance, in the **Topic Migration Job**, the `AuditTopicMigrateJob` simulates all steps of the main `MigrateTopicJob` but does not write changes to the datastore. Similarly, we will follow this paradigm for the `TruncateUserBioJob`.
+
+The objective of our audit job, `AuditTruncateUserBioJob`, is to:
+
+1. Identify user records with bios exceeding 200 characters.  
+2. Simulate truncation logic for these records without saving the changes.  
+3. Provide a detailed report of all affected records, ensuring we are confident in the data to be modified before running the actual migration job.
+
+#### **Thought Process for the Audit Job**
```
seanlip: This isn't a thought process, it's more like a list of factors to consider. A process is more like an ordered series of steps, e.g. applying each of these in turn to the job you're trying to write.

Tutorial-Learn-how-to-make-a-backend-change-that-has-an-impact-on-server-data.md
```diff
@@ -0,0 +1,677 @@
...
+
+1. **Simulating Logic**: The audit job must simulate the exact same steps as the main Beam job to ensure consistency in logic and results.  
+2. **Read-Only Operations**: Unlike the main job, an audit job should not persist any changes to the datastore. This avoids unintended side effects during testing.  
+3. **Detailed Reporting**: The job should generate a detailed report or log indicating the records that require updates. This transparency helps validate the scope and correctness of the job.  
+4. **Reusable Patterns**: Follow established patterns and conventions for audit jobs in the Oppia codebase.
```
seanlip: I think it would help if you gave a bit more detail on what these are.

seanlip (changes requested):
Thanks @Ash-2k3! Sending a partial review, will continue later.

Tutorial-Learn-how-to-make-a-backend-change-that-has-an-impact-on-server-data.md
```diff
@@ -0,0 +1,677 @@
...
+   * **Scenario**: Some `user_bio` fields exceed the 200-character limit.  
+   * **Expected Outcome**: The job should correctly truncate these fields to the maximum allowed length.  
+4. **Complex Case (Multiple Affected Models)**:  
+   * **Scenario**: Multiple users have bios exceeding the character limit.  
+   * **Expected Outcome**: The job should truncate all affected bios and provide a report indicating the number of truncations performed.
+
+By covering these cases, we can ensure the robustness of the Beam job and gain confidence in its behavior across different scenarios.
+
+> [!IMPORTANT]
+> Practice 10: Based on the explanation above, can you write an Audit Job for our use case? Think about how you can simulate the truncation logic while ensuring the job remains read-only and produces detailed reports.
```
seanlip: This is asking the dev to write an audit job but you then follow this question with testing code, which doesn't match the request.

Tutorial-Learn-how-to-make-a-backend-change-that-has-an-impact-on-server-data.md
```diff
@@ -0,0 +1,677 @@
...
+3. **Detailed Reporting**: The job should generate a detailed report or log indicating the records that require updates. This transparency helps validate the scope and correctness of the job.  
+4. **Reusable Patterns**: Follow established patterns and conventions for audit jobs in the Oppia codebase.
+
+> [!IMPORTANT]
+> Practice 9: Based on the above explanation and thought process. Can you write down the Audit Job for our use case.
+
+The `AuditTruncateUserBioJob` is implemented alongside the main Beam job in the `user_bio_truncation_jobs.py` file. Here’s how it can be implemented:
+
+```python
+class AuditTruncateUserBioJob(base_jobs.JobBase):
```
seanlip: One thing we generally want is for the audit and main jobs to use the same functions (rather than duplicating code). Could you give guidance on structuring the jobs to ensure that?

Tutorial-Learn-how-to-make-a-backend-change-that-has-an-impact-on-server-data.md
```diff
@@ -0,0 +1,677 @@
...
+   * For others, enter a `user_bio` that exceeds 200 characters.  
+   * Leave the `user_bio` field empty for additional cases.  
+4. **Run the Job on Populated Data**  
+   * Navigate back to the **Release Coordinator** page and start the job again.  
+   * Observe the behavior and verify whether it matches the expected outcomes for each scenario:  
+     * **Bios within the limit**: Remain unchanged.  
+     * **Bios exceeding the limit**: Are truncated to 200 characters.  
+     * **Empty bios**: Remain unaffected.
+
+### **Conclusion**
```
seanlip, line 659: Can you explain, somewhere in this tutorial, the "sequence of operations"? I.e. they need to do the domain-layer fix to stop new bios > 200 chars getting added. When do they make this fix, when do they run the job on the server, when are things deployed, etc. -- what is the sequence of operations?

I think this is something that probably needs to be explained at the outset of the tutorial, because fixing an issue that affects existing data is an entire process and that process needs to happen in the right order.

Ash-2k3:
Just an update, I will address comments on this PR by Sunday (12 Jan)

Edit - will address by EOW.

Tutorial-Learn-how-to-make-a-backend-change-that-has-an-impact-on-server-data.md
Ash-2k3: Wdyt about "Understanding Preferences Page and Identifying Key Changes" ? I think it provides a clearer sense of when this step is "done" — once the key changes are identified.

Tutorial-Learn-how-to-make-a-backend-change-that-has-an-impact-on-server-data.md
Ash-2k3: Done, Thanks1

Tutorial-Learn-how-to-make-a-backend-change-that-has-an-impact-on-server-data.md
Ash-2k3: Sounds Good, I have replaced it with what you suggested. Thanks!

Tutorial-Learn-how-to-make-a-backend-change-that-has-an-impact-on-server-data.md
Ash-2k3: Thanks for catching this, done!

Tutorial-Learn-how-to-make-a-backend-change-that-has-an-impact-on-server-data.md
Ash-2k3: I think adding it as hint, made more sense. I have made the changes, wdyt ?

Tutorial-Learn-how-to-make-a-backend-change-that-has-an-impact-on-server-data.md
Ash-2k3: I didn't get your comment. Are you referring to the else if block being indented ?

Tutorial-Learn-how-to-make-a-backend-change-that-has-an-impact-on-server-data.md
Ash-2k3: Thanks for providing this link. I skimmed through it and I think it's great to add it here. I have made changes.

Tutorial-Learn-how-to-make-a-backend-change-that-has-an-impact-on-server-data.md
Ash-2k3: I have made the changes. Regarding the formatting of notes, I think it's fine as it is rn for the sake of consistency across other tutorials.

But I do think what you are suggesting will look better. Should we open an issue to track this ? We can first finalise what are the formats we want to follow for practise questions, notes, callouts etc and then apply them to all tutorials. wdyt ?

Tutorial-Learn-how-to-make-a-backend-change-that-has-an-impact-on-server-data.md
Ash-2k3: Done.

Tutorial-Learn-how-to-make-a-backend-change-that-has-an-impact-on-server-data.md
Ash-2k3: Renamed the heading. Do you think it's better now ?

Tutorial-Learn-how-to-make-a-backend-change-that-has-an-impact-on-server-data.md
Ash-2k3: Done.

Tutorial-Learn-how-to-make-a-backend-change-that-has-an-impact-on-server-data.md
Ash-2k3: Ah thanks for catching this, rephrased it.

Tutorial-Learn-how-to-make-a-backend-change-that-has-an-impact-on-server-data.md
Ash-2k3: Ah good catch, this question is supposed to be about tests. I have fixed it, PTAL

Ash-2k3:
Just an update, I have addressed most of the comments. Two are yet remaining, I ll have to think through for addressing them.

Tutorial-Learn-how-to-make-a-backend-change-that-has-an-impact-on-server-data.md
seanlip: Then have the section title be "Identify the Key Changes to Make". "Understand" is woolly.

That said, I wonder if "Identify the parts of the code that need to be changed" is better, or perhaps "Stop the problem from getting worse"? Conceptually the latter is what you are actually doing, and it complements the Beam job. Or maybe "Stop the error from occurring for new data" ... something like that.

Tutorial-Learn-how-to-make-a-backend-change-that-has-an-impact-on-server-data.md
seanlip: Yup seems OK. Thanks.

Tutorial-Learn-how-to-make-a-backend-change-that-has-an-impact-on-server-data.md
seanlip: Look at the difference between the indentation level of this else block and the if block below it. Shouldn't it only differ by 4 spaces?

Tutorial-Learn-how-to-make-a-backend-change-that-has-an-impact-on-server-data.md
seanlip: Please address **all** parts of reviewers' comments when responding. You have missed the second paragraph entirely, AFAICT.

I think you can just apply the "Note" callout if it looks better. I'm not sure we need a long discussion for that. You can file an issue to update it for the other tutorials.

Tutorial-Learn-how-to-make-a-backend-change-that-has-an-impact-on-server-data.md
seanlip: Please use sentence case: "Read everything first." "Process data in steps." "Write everything last."

This helps make clearer that it's a sentence/instruction and not a slogan.

Tutorial-Learn-how-to-make-a-backend-change-that-has-an-impact-on-server-data.md
seanlip: Thanks. Please also change "Audit Job" to "audit job" if you don't capitalize it elsewhere in normal usage.

Tutorial-Learn-how-to-make-a-backend-change-that-has-an-impact-on-server-data.md
seanlip: Yes, I think so. 

One other thought when reading: I think the best practice for audit jobs is to use the `DATASTORE_UPDATES_ALLOWED` property and have the audit/main beam jobs be forced to do exactly the same thing by subclassing -- you can see examples of this in the codebase. If the author follows that then that should make it easier to ensure that the jobs don't deviate from each other. Can you make the tutorial align with this rule? That might also help shorten this section since you just need to explain how to write the job while making appropriate use of that property. (I know there are jobs in the codebase which don't follow this, but we actually need to make them do so -- this is on my list for dev workflow but let's not exacerbate the current problem.)

Tutorial-Learn-how-to-make-a-backend-change-that-has-an-impact-on-server-data.md
seanlip: The update is still pretty vague and doesn't actually give more detail on what these are. See previous comment about DATASTORE_UPDATES_ALLOWED which could probably help you nail this down a lot more.

Tutorial-Learn-how-to-make-a-backend-change-that-has-an-impact-on-server-data.md
seanlip: It looks good, thanks.

seanlip:
Thanks -- I've left replies to the ones you responded to. PTAL.

Tutorial-Learn-how-to-make-a-backend-change-that-has-an-impact-on-server-data.md
Ash-2k3: Done, I have renamed it - Prevent New Data Violations - Identify Code Changes for Bio Length Limit. Thanks

Tutorial-Learn-how-to-make-a-backend-change-that-has-an-impact-on-server-data.md
Ash-2k3: Oh thanks for catching it, I didn't notice it earlier, fixed it now.

Tutorial-Learn-how-to-make-a-backend-change-that-has-an-impact-on-server-data.md
Ash-2k3: Text between two stars is rendered as bold text and between stars it's rendered as Italic font. If you see the preview of this portion, you'll notice it.

I didn't skip it, but I thought it's covered in my second para (where I asked about formatting of notes). I should have been clear but.

Regarding the notes, I would like to get the major comments approved first (The content changes), then I will update the format of the notes (Will do a self review and also make any changes as needed to the format)

Tutorial-Learn-how-to-make-a-backend-change-that-has-an-impact-on-server-data.md
Ash-2k3: Done, but I don't really understand why ending with full stop is better than with semi column.

Tutorial-Learn-how-to-make-a-backend-change-that-has-an-impact-on-server-data.md
Ash-2k3: Done.

Tutorial-Learn-how-to-make-a-backend-change-that-has-an-impact-on-server-data.md
Ash-2k3: Done, I have made some changes to section 2 and section 3.

Tutorial-Learn-how-to-make-a-backend-change-that-has-an-impact-on-server-data.md
Ash-2k3: Thanks, updated this section, PTAL!

Tutorial-Learn-how-to-make-a-backend-change-that-has-an-impact-on-server-data.md
Ash-2k3: Made some changes to the code for that.

Tutorial-Learn-how-to-make-a-backend-change-that-has-an-impact-on-server-data.md
Ash-2k3: Sounds good, but I think mentioning it at the beginning would be better. My reasoning behind this is that folks should be aware of the entire process before even making code changes (If that makes sense).

Ash-2k3:
@seanlip Thanks for the review, I have addressed your comments PTAL.

(Formatting of Notes is not yet down, I am putting it for last once the content gets finalised, I will do a self review and fix any formatting issues.)

Tutorial-Learn-how-to-make-a-backend-change-that-has-an-impact-on-server-data.md
```diff
@@ -4,24 +4,23 @@
...
   - [Section 4: Testing the Beam Job](#section-4-testing-the-beam-job)
   - [Section 5: Run and Validate the Job](#section-5-run-and-validate-the-job)
     - [**Conclusion**](#conclusion)
       - [We Value Your Feedback](#we-value-your-feedback)
 
 # Introduction
 
-In this tutorial, you will learn how to safely implement backend changes that impact server data, an essential skill for any developer working with applications that store data. We’ll cover key concepts like modifying data models, writing and testing Beam Jobs (Audit and Migration), and documenting a reliable launch process. These skills are fundamental for maintaining data integrity, ensuring data consistency, and avoiding disruptions during backend updates.
+In this tutorial, you will learn how to safely implement backend changes that impact server data, an essential skill for any developer working with data-storing applications. We’ll cover key concepts like modifying data models, writing, and testing Beam jobs (Audit and Migration). These skills are fundamental for maintaining data integrity, ensuring data consistency, and avoiding disruptions during backend updates.
```
seanlip: The sentence starting "We'll" seems incorrect ... it says "we'll cover key concepts like (1) modifying data models, (2) writing, and (3) testing Beam jobs". 

"(2) writing" is a general skill and is probably not what you mean to say, but it's what's implied by the comma.

Tutorial-Learn-how-to-make-a-backend-change-that-has-an-impact-on-server-data.md
```diff
@@ -31,11 +30,33 @@ Before you begin, ensure that you have completed the following steps to set up a
...
 
 # Procedure
 
-## Section 1: Navigating and Understanding the Preferences Page
+Before we dive into the implementation, let’s outline the sequence of operations required to enforce and apply the new bio length limit safely. Fixing an issue that affects existing data is a structured process that must follow the correct order:
+
+1. **Modify the Backend Layer to Prevent Future Violations**
+	- Update the backend validation to ensure that new or updated bios cannot exceed 200 characters.
+	- This ensures that, once we fix existing data, no new invalid entries will be introduced.
+2. **Implement the Data Migration and Audit Jobs**
```
seanlip, line 43: For readability, consider having single lines separating each of these numbered points (i.e. add a new line above this line, and similarly below).

Tutorial-Learn-how-to-make-a-backend-change-that-has-an-impact-on-server-data.md
seanlip: It's actually a colon, not a semicolon. 

TBH, I think that specific distinction is more a matter of style -- the "slogan" comment was more for the sentence case vs capital letters. For me, having periods means that one can scan just the bold parts ("Read everything first. Process data in steps. ...") and it's still coherent, whereas a colon means that both parts should be read together. But this specific case is more of a preference and not really a hard-and-fast rule.

seanlip (changes requested):
Hi @Ash-2k3 -- took a pass, on a quick skim it looks good in general!

Could you please update the notes and fix the remaining small issues, then reassign for final review? Thanks.

Tutorial-Learn-how-to-make-a-backend-change-that-has-an-impact-on-server-data.md
Ash-2k3: Ah, the comma shouldn't have been here. Removed it, thanks for catching this!!

Tutorial-Learn-how-to-make-a-backend-change-that-has-an-impact-on-server-data.md
Ash-2k3: SG, Done, thanks!

Ash-2k3:
Thanks for the review @seanlip, I have addressed your comments and also changed the formatting of notes, PTAL, Thanks!

(And also sorry for the delay here.)

Tutorial-Learn-how-to-make-a-backend-change-that-has-an-impact-on-server-data.md
seanlip: Probably more correct to say "We’ll cover key concepts like modifying data models, as well as writing and testing Beam jobs...", since otherwise the first comma needs to be an "and" since you really only have two things here.

seanlip (commented):
One small note, otherwise no concerns. Thanks!

Tutorial-Learn-how-to-make-a-backend-change-that-has-an-impact-on-server-data.md
Ash-2k3: SG, Done!

Ash-2k3:
Addressed the last comment, PTAL @seanlip, thanks!

seanlip (approved):
Thanks! LGTM.

PR #451, opened 2025-04-06
Update codebase overview to use 'domain objects' terminology. Clarify backend structure a bit.

seanlip:
This PR removes the use of "domain models" from the overview doc, since that terminology is not in use and is confusing.

### DIFF
--- a/Overview-of-the-Oppia-codebase.md
+++ b/Overview-of-the-Oppia-codebase.md
@@ -71,8 +71,9 @@ sequenceDiagram
     service->>domain: get_domain_object(datastore_model)
     Note over domain: domain_object = Model(datastore_model)
     domain->>service: domain_object
+    Note over service: Perform any business logic here
     service->>controller: domain_object
-    controller->>web server: JSON or HTML Response
+    controller->>web server: JSON Response
     web server->>client: HTTP Response
 ```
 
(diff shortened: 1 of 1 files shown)


Overview-of-the-Oppia-codebase.md
```diff
@@ -105,31 +103,19 @@ def handler(self, argument_1, argument_2):
...
-
-### Backend models
-
-Before we talk about the domain layer, we need to discuss models.
-
-A model is a data structure, often a class, that stores the information that defines a particular object in Oppia. For example, we have a model for each exploration. (Note that we use "model" to refer to both the class that defines the exploration model and each instance of that class. Sorry, we know it's confusing, but this language is all over the code base.)
-
-There are many different kinds of models in Oppia. We have frontend models, backend (domain) models, and storage models.
-
 ### Domain layer
```
U8NWXD, line 102: the backend architecture diagram shows separate domain and service layers. Should that be changed too?

U8NWXD (changes requested):

Overview-of-the-Oppia-codebase.md
seanlip: Thanks, good catch. They should still be separate, but on closer examination I think the original diagram was a bit wrong, so I simplified it.

seanlip (commented):
@U8NWXD Updated, PTAL. Thanks!

U8NWXD (approved):

PR #473, opened 2025-09-28
Updated steps for locally testing state migrations

reactor0533:
Included link to debugging datastore doc and added clarification for switching to a different branch with the local server running.

Reason for changes:
- When I was working on https://github.com/oppia/oppia/pull/23382, the step 'Checkout the feature branch which contains state migration.' was a bit ambiguous and it took me a few tries to realise that I had to switch to a different branch in another terminal window and that the changes would automatically be detected.
- I think it would be beneficial to mention how to access the local datastore on this page as it makes validating and testing the migration easier, and it helped me figure out what was wrong with my initial migration approach.

### DIFF
--- a/Writing-state-migrations.md
+++ b/Writing-state-migrations.md
@@ -38,7 +38,7 @@ If you find new test files where changes needed to be required, try updating the
 - Start the server and go to the admin page.
 - Load all demo exploration.
 - Create a new exploration, make some changes and save them.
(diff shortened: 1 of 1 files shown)


reactor0533:
Hey @seanlip PTAL, thanks!

seanlip (approved):
Thanks @aroch17 -- looks great! Thanks for making the documentation better :)

PR #479, opened 2025-11-10
Deprecate installing Oppia under Docker

U8NWXD:
There are a number of issues with installing Oppia under Docker (e.g. [#23631](https://github.com/oppia/oppia/discussions/23631), [#23605](https://github.com/oppia/oppia/discussions/23605)), and since we no longer have any active developers using the Docker installation (according to responses to an email to all developers), we are struggling to keep it maintained.

We originally planned to develop exclusively using the Docker installation so that developer environments would be kept consistent, but this ended up being infeasible because for many developers, the overhead of running Docker prevented their systems from running the development server locally. Without this consistency advantage, there isn't much of an advantage to installing under Docker, so maintaining a parallel installation path isn't worthwhile. Note that there are security benefits to sandboxing the development server inside Docker, but since this sandbox would be irrelevant in production, we'd rather devote effort to making Oppia secure both in development and in production. Worried developers can always run the development server inside a virtual machine for similar sandboxing protections.

### DIFF
--- a/Installing-Oppia-using-Docker.md
+++ b/Installing-Oppia-using-Docker.md
@@ -1,5 +1,8 @@
 This guide provides step-by-step instructions for installing Oppia using Docker. Docker simplifies the installation process and ensures a consistent environment across different systems, making it easier to set up and run Oppia.
+
 If you get stuck somewhere and need help, please refer to our [[Getting Help Page|Get-help]].
+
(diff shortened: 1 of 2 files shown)


jayam04 (approved):

jayam04:
@U8NWXD Thanks. LGTM.

PR #484, opened 2025-11-25
Remove all Docker references from the wiki

seanlip:
We no longer support the docker installation, for the following reasons:

- The Docker installation is fully broken right now because it depends on versions of Ubuntu whose package repos are no longer available.
- Our developers do not currently use the Docker installation, which makes it hard for us to troubleshoot when new contributors run into issues with Docker. This undermines the main benefit of using Docker (easier installs for new contributors).
- Supporting Docker will take many developer resources.
- We don't have good support for Docker on CI.
- Many of the dev workflow team members have limited knowledge of Docker.

As a result, the dev workflow team has decided to remove Docker entirely from the Oppia codebase. This PR removes all instances of Docker from the wiki. The corresponding code-removal PR is at https://github.com/oppia/oppia/pull/23936.


seanlip:
@jayam04 @U8NWXD PTAL, thanks!

Google-Summer-of-Code-2024.md
U8NWXD, line 1: should we really be changing the GSoC pages after the fact? I think it's fine to leave references to Docker here since it reflects the state of our development processes at the time. ditto for the 2025 page

U8NWXD (changes requested):

Google-Summer-of-Code-2024.md
seanlip: @U8NWXD This was actually intentional -- I thought the same thing at first, but then I realized that prospective 2026 contributors are likely going to be reading the recent wiki pages for the "general info" (not the project-specific stuff). So I left the Docker stuff in the latter as-is (since posterity is important there), but made a modification to this Q&A.

I think the change is OK since it is still recorded in version history and it doesn't change the spirit of the instructions substantially. Thoughts?

seanlip (commented):
@U8NWXD Responded, PTAL. Thanks!

Google-Summer-of-Code-2024.md
jayam04: @seanlip's approach seems fine to me, as we are only changing FAQs.

jayam04 (approved):
Everything looks good to me. Also, for the @U8NWXD's question, I think @seanlip's reasoning is fine.

U8NWXD (approved):

